PR #7, opened 2023-05-24
Exted ReportObserver class

mirpedrol:


plugins/nf-co2footprint/src/test/nextflow/co2footprint/CO2FootprintFactoryTest.groovy
```diff
@@ -17,21 +17,65 @@
...
+            }
+        }
+        Plugins.init(root, 'dev', manager)
+    }
+
+    def cleanup() {
+        Plugins.stop()
+        PluginExtensionProvider.reset()
+        pluginsMode ? System.setProperty('pf4j.mode',pluginsMode) : System.clearProperty('pf4j.mode')
+    }
```
skrakau, line 69: What's this for?

mirpedrol: These functions are run before (setup) and after (cleanup) tests. I copied them from `CO2FootprintDslTests.groovy` and other plugin test files. We don't need them for this test, but I added it in case we add more tests in the future and becomes useful.

PR #9, opened 2023-06-06
add energy consumption calculation

mirpedrol:
Example output trace file

```
task_id	energy_consumption	CO2e
1	20.328770833333333mWh	9.656166145833334mg
2	20.328770833333333mWh	9.656166145833334mg
3	20.328770833333333mWh	9.656166145833334mg
4	20.328770833333333mWh	9.656166145833334mg

```

Example summary trace file

```
The total CO2 emission is: 38.624664583333335mg
The total energy consumption is: 81.31508333333333uWh
```

### DIFF
--- a/plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintFactory.groovy
+++ b/plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintFactory.groovy
@@ -115,6 +115,31 @@ class CO2FootprintFactory implements TraceObserverFactory {
     }
 
 
+    private String convertToReadableUnits(double value, int unitIndex=4) {
+        def units = ['p', 'n', 'u', 'm', ' ', 'K', 'M', 'G', 'T', 'P', 'E']  // Units: pico, nano, micro, mili, 0, Kilo, Mega, Giga, Tera, Peta, Exa
+        
+        while (value >= 1000 && unitIndex < units.size() - 1) {
+            value /= 1000
+            unitIndex++
+        }
+        while (value <= 1 && unitIndex > 0) {
+            value *= 1000
+            unitIndex--
+        }
+        
+        def formattedValue = value.toString()
+        
+        if (value % 1 == 0) {
+            formattedValue = value.toInteger().toString()
+        } else if (value % 0.1 == 0) {
+            formattedValue = String.format('%.1f', value)
+        } else if (value % 0.01 == 0) {
+            formattedValue = String.format('%.2f', value)
+        }
+        
+        return "${formattedValue}${units[unitIndex]}"
+    }
+
     // Core function to compute CO2 emissions for each task
     List<Double> computeTaskCO2footprint(TraceRecord trace) {
         // C = t * (nc * Pc * uc + nm * Pm) * PUE * CI * 0.001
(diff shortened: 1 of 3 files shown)


plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintFactory.groovy
```diff
@@ -114,8 +115,23 @@ class CO2FootprintFactory implements TraceObserverFactory {
     }
 
 
+    private String convertToReadableUnits(double value, int unitIndex=4) {
+        def units = ['p', 'n', 'u', 'm', ' ', 'K', 'M', 'G', 'T', 'P', 'E']  // Units: pico, nano, micro, mili, 0, Kilo, Mega, Giga, Tera, Peta, Exa
```
skrakau, line 119: like this, if the unit is `gram` no unit is given out. Shouldn't `g` be given out as well?

skrakau (commented):
Hi, this looks already great! I commented on two minor things to add.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintFactory.groovy
skrakau: ah, my mistake :D it's just the prefix

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintFactory.groovy
```diff
@@ -114,8 +115,23 @@ class CO2FootprintFactory implements TraceObserverFactory {
     }
 
 
+    private String convertToReadableUnits(double value, int unitIndex=4) {
```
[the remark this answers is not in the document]
mirpedrol: Thanks for the suggestion! I will try to move this function somewhere where we can use it for different things, I found that I also need it in the CO2Record class for the record plots. 
I will also add it for total values 👍

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintFactory.groovy
mirpedrol: This function outputs only the "prefix", the actual unit should be added afterwards, that way we can reuse it for different units, such as g, Wh...

mirpedrol: exactly, I didn't see your second message before replying 😄

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintFactory.groovy
```diff
@@ -326,13 +348,19 @@ class CO2FootprintFactory implements TraceObserverFactory {
...
+
+
+
+            co2eRecords[taskId] = new CO2Record((Double) eConsumption, (Double) co2, trace.get('name').toString())
+            total_energy += eConsumption
             total_co2 += co2
 
             // save to the file
-            writer.send { PrintWriter it -> it.println("${taskId}\t${co2}"); it.flush() }
+            writer.send { PrintWriter it -> it.println("${taskId}\t${convertToReadableUnits(eConsumption,5)}Wh\t${convertToReadableUnits(co2)}g"); it.flush() }
```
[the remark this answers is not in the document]
mirpedrol: I think it's good to let the user decide which fields to output :) but maybe out of the scope of this PR

skrakau:
Great, LGTM!

skrakau (approved):

PR #11, opened 2023-06-07
Add first test for CO2 calculation

skrakau:
Here comes a first simple test for the CO2 calculation: comparing the CO2 value against the result from http://calculator.green-algorithms.org/


mirpedrol (commented):
I think it looks good, but should we try to add the changes to the Nextflow trace record before merging? otherwise tests will be failing

mirpedrol (approved):
let's do it 😄

PR #12, opened 2023-06-13
Update Report - Add energy consumption plots and add values to the table

mirpedrol:
Follow up from #9 
Add energy consumption plot

New plots (conditional tickformat):
image
Example table with converted units:
image
Example table with raw values:
image

### DIFF
--- a/plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintReportSummary.groovy
+++ b/plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintReportSummary.groovy
@@ -195,16 +195,16 @@ class CO2FootprintReportSummary {
             final result = new LinkedHashMap<String,?>(12)
             final sorted = tasks.sort( false, { CO2Record co2record -> metric.call(co2record) } )
 
-            result.mean = round(total / count as double)
-            result.min = round(quantile(sorted, 0))
-            result.q1 = round(quantile(sorted, 25))
-            result.q2 = round(quantile(sorted, 50))
-            result.q3 = round(quantile(sorted, 75))
-            result.max = round(quantile(sorted, 100))
+            result.mean = total / count as double
+            result.min = quantile(sorted, 0)
+            result.q1 = quantile(sorted, 25)
+            result.q2 = quantile(sorted, 50)
+            result.q3 = quantile(sorted, 75)
+            result.max = quantile(sorted, 100)
 
             // discard entry with all zero 
-            if( result.min == 0 && result.min == result.max  )
-                return null
+            //if( result.min == 0 && result.min == result.max  )
+            //    return null
 
             result.minLabel = minLabel
             result.maxLabel = maxLabel
--- a/plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintReportSummary.groovy
+++ b/plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintReportSummary.groovy
@@ -202,9 +202,12 @@ class CO2FootprintReportSummary {
             result.q3 = quantile(sorted, 75)
             result.max = quantile(sorted, 100)
 
-            // discard entry with all zero 
-            //if( result.min == 0 && result.min == result.max  )
-            //    return null
+            /* 
+                Unlike the Nextflow Report, we are not rounding the results nor discarding entries with all zeros.
+                This decision is taken to avoid the loss of information in the report. 
+                Plots will show values of 0, making the representation of all processes consistent.
+                This class reports all values in mili-unit to increase precision. Values are converted to the required units by CO2FootprintReportTemplate.js
+            */           
 
             result.minLabel = minLabel
             result.maxLabel = maxLabel
--- a/plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintReportSummary.groovy
+++ b/plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintReportSummary.groovy
@@ -35,12 +35,13 @@ class CO2FootprintReportSummary {
 
     static {
         mappers.co2e = { CO2Record co2record -> co2record.getCO2e() as Double }
+        mappers.energy = { CO2Record co2record -> co2record.getEnergyConsumption() as Double }
     }
 
 
 
     /**
-     * Hold the summary for each series ie. co2e (only currently)
+     * Hold the summary for each series ie. co2e
      */
     private Map<String, Summary> series
 
@@ -57,7 +58,7 @@ class CO2FootprintReportSummary {
     }
 
     /**
-     * @return The list of series names eg `co2e` (only currently)
+     * @return The list of series names eg `co2e`
      */
     List<String> getNames() { names }
 
@@ -77,7 +78,7 @@ class CO2FootprintReportSummary {
      * Compute the summary stats for the collected tasks
      *
      * @param
-     *      name A series name eg {@code co2e} (only currently)
+     *      name A series name eg {@code co2e} 
      * @return
      *      A {@link Map} holding the summary containing the following stats:
      *      - min: minimal value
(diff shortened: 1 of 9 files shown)


mirpedrol:
@skrakau I have added a function to convert the values to more readable units, adding kilo, mega, mili... for the trace files (see example in https://github.com/qbic-pipelines/nf-co2footprint/pull/9), do we want to do the same for plots in the report? I think it can be confusing if we output a different unit every time if people want to compare different runs, but on the other hand if values are too close to 0 we won't be able to see the differences.

skrakau:
Hi @mirpedrol ,
for the memory plots in the standard execution report, for example, the units are adjusted for each run as well, so this shouldn't be a problem as far as I see.

mirpedrol:
I am unsure about rounding the values, they are now rounded for the "Raw" plots. Should we also do it for "Readable Units" in the table?

skrakau:
One plot with adjusted y axis labels should be sufficient.

skrakau:
Some notes already regarding the tasks table:
- the should be a white space in front of the unit (for Human readable)
- it might be helpful to display the unit for the `raw values`, e.g. `(gram)`, as in https://www.nextflow.io/docs/latest/tracing.html#tasks

skrakau (approved):
LGTM!

skrakau:
A few notes
- regarding the currently generated CO2 reports: the plots themselves seem fine, just the y axis tick labels (and the displayed boxplot labels/quantiles) seem not OK. Or do I miss something here?
- why are the y axis labels like they are, i.e. [0.0, 0.0, ...., 0.0] ?

would the following help?
- the rounding of the quantile values in the 'CO2F*ReportSummary.groovy' mainly seem to cause a problem in combination with discarding zeros and because the stored unit might be not optimal (g, but the amounts for such super fast tasks are more in the range of milligrams or milliwatt-hours). If we could internally store the vales as milligram and milliwatt-hours as milliseconds are used for the time by nextflow, then the rounding wouldn't cause a problem anymore (if it would still make sense to round, would tasks with < 0.005 mg CO2 emission occur?)
- a rounding in `CO2F*ReportTemplate.js` would not be needed, right?
- (if one would use milliwatt-hours maybe it would make sense to write the full name of the units to the task table header for the raw values to avoid confusions with MWh?)
- remains the issue with the y axis labels for small values: https://plotly.com/javascript/tick-formatting/#using-exponentformat maybe this could help?
- consider: tickformat: '.1f' -> tickformat: '.4s'
- a remaining question is, if one would want to adjust the y axis tick labels as it is done for the Nextflow Memory plot (no unit given in the y axis label/title, but the tick label says 12T ...) . Or if one would do it as it is done for the Execution time (minutes) , which is always given in minutes, if I see it correctly.

mirpedrol:
Steps after in person discussions:

1. Convert all values to mili (raw) and round. For the table convert to readable units.
2. Display lower values with scientific notation in the y axis. (check after using all values in mili)
3. Also modify the tickformat. (maybe)
4. Check if we can add the unit in the plot ticks, what happens if we put miligram instead of gram?

mirpedrol:
If the maximum value is 4 or lower than 4, we use a tickformat of .2f (displays the values as decimals 0.5, etc), if it's higher than 4 (or some other higher reasonable number) we use .2s (adds the units to the tick value 1k, etc). 
This way we can be sure that for lower values we will never have a tick with 500m, as the maximum number of displayed ticks is 5.

skrakau:
4 because, the default `nticks` value is 5, leading to max. 5 ticks. Thus with values >= 4, the min. tick is at >= 1.0 (g).

plugins/nf-co2footprint/src/resources/assets/CO2FootprintReportTemplate.js
```diff
@@ -77,8 +76,25 @@ $(function() {
...
+      max = Math.max(max, p.y);
+    }
+  });
+  var co2e_tickformat = (max <= 4) ? ('.2f') : ('.2s');
+  energy_data.forEach(function (p) {
+    max = 0;
+    if (p !== null) {
+      max = Math.max(max, p.y);
+    }
+  });
```
skrakau: Hi, when checking the output I observed that for some low values it still uses the `.2s` `tickformat`. I briefly checked the code: `max` is `NaN` (i.e. that's why the tickformat always stays `.2s`) and it seems that something is wrong with the max() assignment.

mirpedrol: Solved now, the problem happened when there are boxes in the plot, then `p.y` is an array

skrakau: Great, the plot for the smaller values looks good now!

skrakau (commented):
For the standard execution report they use `.4s` (and `.1f` for time etc.).

If I see it right, in this way only for values < 10 the decimal places are the same, for everything above they are different anyway. Did you test it for a larger pipeline run? (But we probably first need to find one creating really large values...) But couldn't we then keep it consistent with the other execution plots and use `.4s`?

skrakau:
If you think it's better this way, we can also keep it like this

mirpedrol:
Using `.1f` sounds good but I don't like `.4s` because I think it's confusing with thousands. 
image
I can't find a pipeline to test with big numbers but the exmaple plots look like this (values around 3 milion)
image

skrakau:
I see your point, I guess the reason for the execution plots is that this it allows more accurate values, in particular for the displayed boxplot values

skrakau (approved):
LGTM now :)

plugins/nf-co2footprint/src/resources/assets/CO2FootprintReportTemplate.js
```diff
@@ -1,21 +1,14 @@
 // JavaScript used to power the Nextflow Report Template output.
 window.data_byprocess = {};
 
-/* helper functions that takes an array of numbers each of each
-   is a integer representing a number of bytes and normalise to base 2 scale */
-function norm_mem( list ) {
+/* helper functions that takes an array of numbers 
+    units are in miliWh or milig and are converted to its base unit */
```
skrakau, line 5:
```suggestion
    units are in milliwatt-hours (mWh) or milligramm (mg) and are converted to its base unit */
```

plugins/nf-co2footprint/src/resources/CO2FootprintReportTemplate.html
```diff
@@ -203,24 +203,44 @@ <h4>Workflow execution completed unsuccessfully!</h4>
...
+    <h2 id="resources" style="padding-top: 80px;">CO2 Footprint Measures</h2>
     <p>These plots give an overview of the distribution of resource usage for each process.</p>
 
     <h4>CO2e</h4>
     <ul class="nav nav-tabs" id="co2eplot_tabs" role="tablist">
       <li class="nav-item">
-        <a class="nav-link active" id="co2eplot_tablink" data-toggle="tab" href="#co2eplot_tabpanel" role="tab" aria-controls="pctco2eplot_tabpanel" aria-expanded="false">
-          Raw Usage
+        <a class="nav-link active" id="co2eplot_tablink" data-toggle="tab" href="#co2eplot_tabpanel" role="tab" aria-controls="co2eplot_tabpanel" aria-expanded="false">
+          Raw Emisions
```
skrakau, line 213:
```suggestion
          Raw Emissions
```

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintReportSummary.groovy
```diff
@@ -194,16 +195,16 @@ class CO2FootprintReportSummary {
...
-            result.q3 = round(quantile(sorted, 75))
-            result.max = round(quantile(sorted, 100))
+            result.mean = total / count as double
+            result.min = quantile(sorted, 0)
+            result.q1 = quantile(sorted, 25)
+            result.q2 = quantile(sorted, 50)
+            result.q3 = quantile(sorted, 75)
+            result.max = quantile(sorted, 100)
 
             // discard entry with all zero
```
mirpedrol, line 205: write a comment explaining why we don't round and why we don't discard zeros

PR #13, opened 2023-06-19
Add CO2 formula values to Report

mirpedrol:
Follow up from #12 

example output:
image
All extra calculation values are added as "metadata".


mirpedrol:
Closing in favour of https://github.com/qbic-pipelines/nf-co2footprint/pull/28

PR #18, opened 2023-07-28
Handle powerdraw memory parameter

skrakau:
Added powerdraw memory parameter to config


mirpedrol (approved):
LGTM!
Should we also add a test for this value like with PUE?

skrakau:
Thanks! I thought maybe we can add a test with a combination of other parameter a bit later, to reduce the number of tests a bit

skrakau:
Ah, I forgot to mention, thememory powerdraw parameter is not a parameter that that can be adjusted on http://calculator.green-algorithms.org/, that's why we can not test it against the result from the website (but can still add tests again our own results of course)

PR #19, opened 2023-07-31
Convert test results units for comparison

skrakau:
Convert units and number of decimal places of the results in tests again for comparison to results from www.green-algorithms.org

I thought we compare the values here to the values from the green algorithm project (for now, as long as we don't change anything in the calculation at least), to avoid that we include errors from our side.


mirpedrol (approved):
LGTM

PR #21, opened 2023-08-02
Remove CO2FootprintExtension code to avoid generation of output files when running tests

skrakau:
Plugin output files were generated and not removed each time when running plugin tests. This was caused (only) by tests in `CO2FootprintDslTest.groovy`.

Removed unused code from nf-hello template: 
- `CO2FootprintExtension.groovy`
- corresponding tests in `CO2FootprintDslTest.groovy`


mirpedrol (approved):
🚀

PR #23, opened 2023-08-03
Add report file name to config

skrakau:
Add report file name to config and changed default name to lower case for consistency reasosn.


mirpedrol (approved):
LGTM

PR #25, opened 2023-08-04
Add further task-specific params to TXT output file

skrakau:
- I changed `def Double` to `Double`, because `def` is for variables of unknown type
- added the remaining task-specific params to the `CO2Record` class and added helper function for conversions
- changed some `Float`s to `Double`s to avoid conversion artefacts when converting them to `Double`s (when adding to the list). I would suggest we systematically change the structure afterwards (not assigning those floats to `List<Double>` anymore) and adjust all types to floats where possible. 

I just added the parameters, we should think about restructuring and cleaning up a bit in the near future, when the content is more complete.

### DIFF
--- a/plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/HelperFunctions.groovy
+++ b/plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/HelperFunctions.groovy
@@ -13,7 +13,7 @@ public class HelperFunctions {
             value *= 1000
             unitIndex--
         }
-        
+        value = Math.round( value * 100 ) / 100
         return "${value} ${units[unitIndex]}"
     }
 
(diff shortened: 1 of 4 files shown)


plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/HelperFunctions.groovy
```diff
@@ -13,8 +13,37 @@ public class HelperFunctions {
...
             unitIndex--
         }
-        
-        return "${value}${units[unitIndex]}"
+        value = Math.round( value * 100 ) / 100
+        return "${value} ${units[unitIndex]}"
+    }
+
+    static public String convertBytesToReadableUnits(double value) {
+        def units = ['B', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB']  // Units: pico, nano, micro, milli, 0, Kilo, Mega, Giga, Tera, Peta, Exa
```
mirpedrol, line 21:
```suggestion
        def units = ['B', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB']  // Units: Byte, Kilobyte, Megabyte, Gigabyte, Terabyte, Petabyte, Exabyte
```

mirpedrol (approved):
LGTM!

PR #28, opened 2023-08-08
Add task params to report

mirpedrol:
Add task specific params to the report table.
- Added time, cpus, powerdrawCPU, cpuUsage and memory
- Modify JS unit conversion functions to work as the groovy helper functions

Example (raw values)
image
Example (human readable)
image
- I classified them as metadata, as metrics I left CO2 emissions and energy consumption which are the final metrics

### DIFF
--- a/plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2Record.groovy
+++ b/plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2Record.groovy
@@ -39,7 +39,7 @@ class CO2Record extends TraceRecord {
                 'energy':           energy,
                 'co2e':             co2e,
                 'time':             time,
-                'co2e':             co2e,
+                'cpus':             cpus,
                 'powerdrawCPU':     powerdrawCPU,
                 'cpuUsage':         cpuUsage,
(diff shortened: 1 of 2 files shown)


plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2Record.groovy
```diff
@@ -30,7 +30,6 @@ class CO2Record extends TraceRecord {
         this.energy = energy
         this.co2e = co2e
         this.time = time
-        this.cpus = cpus
         this.powerdrawCPU = powerdrawCPU
         this.cpuUsage = cpuUsage
```
skrakau, line 34: Why is this removed? 
It seems I forgot to add it to `this.store` below, but wouldn't it be needed also for the report table?

mirpedrol: Ups that was unintentional, I only wanted to remove the duplicated co2e line below, thanks for spotting it!

skrakau (approved):
Great, LGTM :)

PR #30, opened 2023-08-14
add total value of co2 and energy consumption to Report

mirpedrol:
image


skrakau (approved):
Great, looks good! I removed a whitespace

PR #31, opened 2023-08-14
Add equivalences to report

mirpedrol:
image

### DIFF
--- a/plugins/nf-co2footprint/src/resources/CO2FootprintReportTemplate.html
+++ b/plugins/nf-co2footprint/src/resources/CO2FootprintReportTemplate.html
@@ -21,6 +21,7 @@
   <meta name="description" content="Nextflow workflow report for run id [${workflow.runName}]">
   <meta name="author" content="Sabrina Krakau">
   <link rel="icon" type="image/png" href="https://www.nextflow.io/img/favicon.png" />
+  <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">
 
   <title>[${workflow.runName}] Nextflow Workflow Report</title>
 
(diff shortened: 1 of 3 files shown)


skrakau:
Let's postpone the tests. I will first create an extra PR with a test for the total CO2 value, which is not covered yet. Based on thats we can then also add tests for the equivalences.

plugins/nf-co2footprint/src/resources/CO2FootprintReportTemplate.html
```diff
@@ -199,6 +200,43 @@ <h4>Workflow execution completed unsuccessfully!</h4>
...
+
+      <p>This is equivalent to:</p>
+      <dl class="row small">
+        <dt class="col-sm-6"><i class="fa fa-tree"></i> Months a tree would need to sequester this amount of CO2</dt>
+        <dd class="col-sm-3">${co2_totals.tree}</dd>
+
+        <dt class="col-sm-6"><i class="fa fa-car"></i> Km by car</dt>
+        <dd class="col-sm-3">${co2_totals.car}</dd>
+
+        <dt class="col-sm-6"><i class="fa fa-plane"></i> Km by plane</dt>
```
skrakau, line 221: I would tend to stick to only one plane equivalence value.

skrakau:
Very nice! :)
Maybe we could even think about highlighting the total values in a separate section a bit more (not within the grey area).

plugins/nf-co2footprint/src/resources/CO2FootprintReportTemplate.html
mirpedrol: Sounds good, should we keep the % of a flight Paris-London to have a different metrics? As we already have km by car

skrakau (approved):
LGTM! :)

PR #38, opened 2023-10-04
make the emojis nicer

mirpedrol:
Making the emojis colorful. We can use regular emojis, they should be available in all browsers (tested with chrome & safari). The other option is to still use the library font-awesome and use solid colors.

image

image

For reference, [here](https://www.w3schools.com/charsets/ref_emoji_smileys.asp) I found all the decimal codes for emojis


skrakau:
Actually I tend to prefer the `font-awesome` option, which is a bit more minimalistic and it would look the same across all browsers. What would be your preference? 

Maybe the sentences on the right side of it could be aligned?

ewels:
Agree with @skrakau that I prefer the monochrome / minimalistic approach. But I also agree with @mirpedrol that it'd be nice to remove the dependency on Font Awesome.

I think that the best approach would be to use inline SVGs. You can find loads here: https://icon-sets.iconify.design/

No dependencies and many to choose from.

mirpedrol:
Addressed in https://github.com/nextflow-io/nf-co2footprint/pull/47

PR #39, opened 2023-10-04
add 'cpu_usage is null' message to debug instead of warning

mirpedrol:
This message is printed for every task, polluting the pipeline execution output in the terminal. Here I print them as a debug message instead. The message will be printed to `.nextflow.log`


skrakau (approved):
Great!

PR #40, opened 2023-10-04
remove redundant information from output report

mirpedrol:
remove redundant information from output report, it's available in the regular NF report

image


skrakau (approved):
LGTM!

PR #44, opened 2023-10-05
add energy consumption plot to a tab

mirpedrol:
The energy consumption plot doesn't occupy the whole page space, still don't know why

image
image


skrakau:
hmm... I also don't have an idea on the first view

PR #45, opened 2023-10-05
Add co2e description

skrakau:
I added a description for CO2e. Not sure about the location, maybe we can move it in the future to another location in hte report, just thought it might be good if it's already there.


mirpedrol (approved):
LGTM!
I like this location, then we can add the equation below this text :)

PR #59, opened 2023-10-17
Mention not working for GPUs yet

skrakau:
Mention not working for GPUs yet


mirpedrol (approved):
Only a small suggestion, we could add an admonition such as `!!! warning`

PR #64, opened 2023-10-17
Add metrics to task output and html task table

skrakau:
- added task `name`, `status` and `cpu_model` to txt task output
- added `cpu_model` to `CO2Record`
- fixed formatting: due to placement of `,`, `FIELDS` entries after `name` caused an error in the html output, because two `,`s. 
- added `cpu_model` to HTML report task table


mirpedrol (approved):
LGTM

PR #66, opened 2023-11-07
Allow custom CPU TDP input file

skrakau:
Allow custom CPU TDP input file and added description to docs.

I decided in the end to allow for a CSV file, since with this it maybe easier to collect the data for some use cases.


skrakau:
> LGTM! It would be good to also add a test for this :)

True, but not sure how to (quickly) add a test using an input file, will add this to the list.

PR #68, opened 2023-11-07
Describe also summaryFile parameter

skrakau:
Describe also summaryFile parameter.

This output file was not explicitly mentioned yet, since I wasn't sure if we would keep it, but since it is anyway generated I mentioned it now in case someone want to change it.


mirpedrol (approved):
LGTM!

Issue #58, opened 2023-10-17
Add parameter to disable use of cpu_model 


skrakau:
Add parameter to disable use of `cpu_model` and use default power draw or provided CPU core power draw value.  Add parameter for custom power draw value.

skrakau:
Solved in #70

PR #70, opened 2023-12-19
Add parameter to ignore cpu_model and to change default CPU power draw value

skrakau:
I added two parameters
- `ignoreCpuModel`: to ignore the retrieved `cpu_model` name and use the default CPU power draw value
- `powerdrawCpuDefault`: to change the default power draw value of a computing core. Used in combination with `ignoreCpuModel` or for tasks where no TDP value could be retrieved for the `cpu_model` (either because this is `null` or because there is no entry for the `cpu_model`)

Additionally, i removed the `cpu_model` column from the TXT report if `ignoreCpuModel` is set, since this would be misleading. If the `cpu_model` was not found in the TDP data and the default power draw value is used instead, the `cpu_model` column is kept (would be task specific). In this case, anyway a warning is given out already.

Probably it would make sense, if `ignoreCpuModel` is set, to remove the `cpu_model` column from the table in the HTML report as well. @mirpedrol what do you think and is this something you could add?

Addresses https://github.com/nextflow-io/nf-co2footprint/issues/58

### DIFF
--- a/docs/usage/parameters.md
+++ b/docs/usage/parameters.md
@@ -39,3 +39,5 @@ Mutually exclusive with the `ci` parameter.
 This should contain the following columns: `model`,`TDP`,`n_cores`,`TDP_per_core`.
 Note that this overwrites TDP values for already provided CPU models.
 You can find the by default used TDP data [here](../../plugins/nf-co2footprint/src/resources/TDP_cpu.v2.2.csv).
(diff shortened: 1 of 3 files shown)


mirpedrol:
`cpu_model` values appear as empty (`-`) in the report if `ignoreCpuModel` is used
image

mirpedrol (approved):

PR #72, opened 2023-12-19
Add info about cached processes to docs

skrakau:
Added info that for the total CO2e of a pipeline all completed and cached processes are used. 

@mirpedrol Actually, I usually also use "CO<sub>2</sub>e" for the equation, in the slides at least (just realizing that it's different in the code), could we change this for the formula in the docs as well (currently using "C")?


mirpedrol (approved):

mirpedrol:
I modified the formula to use CO<sub>2</sub>e

PR #74, opened 2023-12-20
replace katex bymathjax to fix rendering formula

mirpedrol:
Currently, the formula is only rendered on reload. This will render the formula on the first click to the subpage.


skrakau (approved):
Great :) LGTM!

PR #81, opened 2023-12-21
collect warning messages and filter out duplicated

mirpedrol:
This is to avoid having the same warning printed for every process. Collects all warning messages and filters out the duplicated ones.

image


skrakau:
I would add an extra warning if the `cpu_model` is `null`, since this is a different issue one should be aware of.

skrakau:
Great, I would still wait a bit to make sure there is no other, intended way of handling redundant warnings

skrakau:
I added some changes, have a look if you think that make sense or if you would want soemthing to be changed

skrakau (approved):
LGTM :)

mirpedrol:
I agree with your changes :) is it ready to be merged?

PR #83, opened 2024-01-11
Renamed output files

skrakau:
Hi, I thought some more consistent output filenames would be nice.

What would you think about
- `co2footprint-<timestamp>.txt` -> `co2footprint_trace_<timestamp>.txt`
- `co2footprint-<timestamp>.summary.txt` -> `co2footprint_summary_<timestamp>.txt`
- `co2footprint-report-<timestamp>.html` -> `co2footprint_report_<timestamp>.html` 
?

Would it make sense to make use of the Nextflow `trace` wording? Then it would be somehow analogous to the nf-core `execution_report_*` and `execution_trace_*` naming.

If yes, then I would also rename the variable names to avoid confusion

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,5 +11,6 @@
 build
 work
 out
-co2footprint-*.txt
+co2footprint_*.txt
+co2footprint_*.html
 site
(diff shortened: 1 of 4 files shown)


skrakau:
OK, ready for review. I also added using a timestamp to the example custom config file.

mirpedrol (approved):
LGTM!

PR #85, opened 2024-01-12
Mention in docs failed tasks will be included

skrakau:
Failed tasks are included as well, I thought it should be mentioned.


mirpedrol (approved):
LGTM

PR #86, opened 2024-01-30
Report nf-co2footprint version

skrakau:
I added the reporting of the nf-co2footprint version. 

Or would there be a more elegant way for this?


mirpedrol (approved):
LGTM!

We have the changelog now to add the changed :) https://github.com/nextflow-io/nf-co2footprint/blob/master/CHANGELOG.md

PR #89, opened 2024-02-05
Report nf-co2footprint parameters 

skrakau:
I added the reporting of the nf-co2footprint params to the summery TXT output file. 

To avoid having to list each individual param for the actual report (prone to being forgotten), I just grouped them by rough topics (input files, output files, params for the actual CO2 calculation) and then used a sorted map. Let me know if you know a more elegant solution.

However, I am not an HTML expert and don't know how to use this for the HTML report generation as well (without having to list each param separately). @mirpedrol would you know how to do that?


skrakau:
Note: first add values manually to HTML report, automize it later for map.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintFactory.groovy
```diff
@@ -737,6 +742,30 @@ class CO2FootprintFactory implements TraceObserverFactory {
...
+            
+            // Render JSON
+            final QUOTE = '"'
+            for( int i=0; i<fields.size(); i++ ) {
+                if(i) result << ','
+                String name = fields[i]
+                String value = all_options[name].toString()
+                result << "{" << QUOTE << "option" << QUOTE << ":" << QUOTE << name << QUOTE << ","
+                result << QUOTE << "value" << QUOTE << ":" << QUOTE << value << QUOTE << "}"
+                //result << QUOTE << name << QUOTE << ":" << QUOTE << value << QUOTE
```
mirpedrol, line 762:
```suggestion
```
This can be deleted :)

mirpedrol (approved):
LGTM!

PR #91, opened 2024-09-11
Align to latest Nextflow version and bump to groovy 4.0

tillenglert:
This PR will align the plugin to the latest nextflow version (24.08.0-edge).

It will deprecate all nextflow versions prior to 24.01.0-edge.

(I might need to add some updates to the test workflows, but we'll see if that can be changed apart of this PR)


plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintFactory.groovy
```diff
@@ -182,7 +182,7 @@ class CO2FootprintFactory implements TraceObserverFactory {
          * Factors of memory power usage
          */
         // nm: size of memory available [GB] -> requested memory
-        Long memory = trace.get('memory') as Long
```
skrakau, line 185: Why did you change this? I think `Long` would be needed here

tillenglert: The function returns a `List<Double>`, and within this list the memory. So to return it actually as said list the memory needs to be cast to double. If the memory is returned as memory, the function would need to be set to return something like List<numerical> which in turn would need casting to doubles again. So in this case I would say the lesser evil is to cast the memory to double, even though it is a non float number.

skrakau: ok, I don't get the exact problem. Here it's also accessed like this: https://github.com/nextflow-io/nextflow/blob/949181053afce9e66ebde1ccff3c725fb2f61697/modules/nextflow/src/main/groovy/nextflow/trace/ReportSummary.groovy#L52

skrakau: (it was about this list: https://github.com/nextflow-io/nf-co2footprint/blob/c01bbde1a8b5ce30ffe8371e9c795f4771f1bff5/plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintFactory.groovy#L219 )

skrakau: conclusion: although `long` would be more optimal, using `double` should be sufficient to represent the memory values here

skrakau: Maybe we can convert the `list` to a `map` in a follow-up PR

tillenglert: I added a comment and a todo to the cast, which is now later to calculate the `nm` and respective used energy with the full precision `memory`

tillenglert:
The force push was to sign off every commit using, which is now required.

skrakau (approved):
LGTM!

PR #94, opened 2024-11-28
update java and grade version

Bastian-Eisenmann:


tillenglert:
Hi @Bastian-Eisenmann, usually java tests should also run on smaller java versions, maybe you can have a look if gradle 8.11 has any requirements which are not met? If not, you can als try to increment the CI tests to other versions, or at least the higher version to java 23 👍

Bastian-Eisenmann:
Hi,
in my previous project we tried to keep the java & tools versions as high as possible (often also with some kind of ci/cd automatic update) and also parallel at all parts of the project.
On my pc was java version 23 installed which hasn´t worked with the given config because gradle 8.9 is not compatible to java 23 (https://docs.gradle.org/current/userguide/compatibility.html) - however, with installing a lower java version the current configuration does probably work.
I can also undo the commit [3c447e2](https://github.com/nextflow-io/nf-co2footprint/pull/94/commits/3c447e27562dd70162362fad5e76fc91ef639718) but would keep the gradle version update so that there will be no problems when java version 23 is used.

tillenglert:
If the version work together, I'm fine with upgrading! I aligned the versions to nextflow, but also they incremented to gradle 8.10 since then! Do you know if grade is backwards compatible with older java versions?

tillenglert (approved):
Great it's working now! Can you sign your commits? There are some code snippets you can use in the details of the CI test 👍

Bastian-Eisenmann:
> That did not work, have you followed the instructions in the details of the CI test? Would you mind opening a new branch and PR so we do not blow up the commits for a version bump? 😆

I did but seems like the final commit was again without that signing message... yeah, I will just create a new one, that should finally work :-)

Issue #26, opened 2023-08-07
Sorting in report table does not account for units


skrakau:
The sorting in the report table does currently not account for units, e.g. 998 mg is considered > 9g.

PR #95, opened 2024-12-03
Fix sorting of human-readable values within the html report summary table

tillenglert:
This PR solves #26 . The issue was the parsing of values when sorting is requested. Also I refactored the variable names within the table and added a rounding function (only in human-readable table) for the core usage factor.


skrakau (approved):
LGTM!

PR #97, opened 2024-12-10
add co2 equivalences in text report

Bastian-Eisenmann:


plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintFactory.groovy
```diff
@@ -359,6 +359,28 @@ class CO2FootprintFactory implements TraceObserverFactory {
             co2eSummaryFile.println("Total CO2e footprint measures of this workflow run")
             co2eSummaryFile.println("CO2e emissions: ${HelperFunctions.convertToReadableUnits(total_co2,3)}g")
             co2eSummaryFile.println("Energy consumption: ${HelperFunctions.convertToReadableUnits(total_energy,3)}Wh")
+
+            List equivalences = computeCO2footprintEquivalences()
+            List<GString> readableEquivalences = new ArrayList<GString>();
+            if (equivalences[0]){
```
tillenglert, line 365: I think cars and trees are always computed, is there a reason you would escape them?

tillenglert (approved):
Apart from my comment, looks good!

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintFactory.groovy
Bastian-Eisenmann: Not for real, just did it like that in the case if somebody changes the downer implementation the logic should still work because the values can be nullable

PR #109, opened 2025-02-19
Add Xeon E5-2670 v3 TDP

LouisLeNezet:
I've added all the cpu missing from my cluster in the v2.2 table.
I've also create a v2.3 table with all the additional data from the http://cpudb.stanford.edu/


plugins/nf-co2footprint/src/resources/TDP_cpu.v2.2.csv
JosuaCarl, line 1: You can provide a file with your specific processors via your config:
```config
plugins {
  id 'nf-co2footprint@1.0.0-beta1'
}

co2footprint {
  customCpuTdpFile = "/path/to/your/custom/CPU_TDPs.csv"
}
```

We try to keep this file unchanged as it is specifically copied form the Green Algorithms project.

plugins/nf-co2footprint/src/resources/TDP_cpu.v2.3.csv
JosuaCarl, line 1: Exactly how it should be done with the current template 👍 . I will check whether all models are present in the newly created dataset (https://github.com/qbic-projects/Green_algorithms_assets/blob/main/cpu_info/specifications_filtered.csv) and integrate missing ones.

plugins/nf-co2footprint/src/resources/cpudb/cpu.R
JosuaCarl, line 1: What does it do ? Just in general.

plugins/nf-co2footprint/src/resources/cpudb/manufacturer.csv
JosuaCarl, line 1: This is very interesting to me ? Where did you get the info from ? You could maybe commit this here: https://github.com/qbic-projects/Green_algorithms_assets/tree/main via a new PR.

plugins/nf-co2footprint/src/resources/TDP_cpu.v2.2.csv
LouisLeNezet: That does make sense

plugins/nf-co2footprint/src/resources/cpudb/cpu.R
LouisLeNezet: It downloads the data from the `cpudb` initiative, aggregate the data from the different tables and update the current `TDP_cpu.v**.csv` with the additional data.

plugins/nf-co2footprint/src/resources/cpudb/manufacturer.csv
LouisLeNezet: This is from the `cpudb` initiative.
http://cpudb.stanford.edu/
But you seem to now have quite a lot more cpu in your dataset now, so I'm not sure if you need this dataset now.

JosuaCarl: I will check for all your CPUs in the new set and append if needed. Thank you for the information ❤️

JosuaCarl:
I close this, because I can confirm that (almost) all models present in the new 2.3 dataset are included into the expanded. The exception here are the following 29 models: 
```python
['A8-7680',
 'A9-9425 SoC',
 'AMD 7552',
 'Any',
 'Core 2 Quad Q6600',
 'Core i3-10350K',
 'Core i9-10900XE',
 'Core i9-10920XE',
 'Xeon E5-2660 v3',
 'Xeon E5-2665',
 'Xeon E5-2670',
 'Xeon E5-2670 v2',
 'Xeon E5-2680 v3',
 'Xeon E5-2683 v4',
 'Xeon E5-2690 v2',
 'Xeon E5-2690 v3',
 'Xeon E5-2695 v4',
 'Xeon E5-2697 v4',
 'Xeon E5-2699 v3',
 'Xeon E5-2699 v4',
 'Xeon E5-4610 v4',
 'Xeon E5-4620',
 'Xeon E5-4650L',
 'Xeon E7-8867 v3',
 'Xeon E7-8880 v4',
 'Xeon L5640 ',
 'Xeon Phi 5110P',
 'Xeon X3430',
 'Xeon X5660']
```
All except `Any` (which is not `default`), the X-Series `[ 'Core i9-10900XE', 'Core i9-10920XE']`
and `'Core i3-10350K'` (which was pulled from the marked by Intel and can only be found in benchmarking datasets that had access to it) can be attibuted to naming changes, which are addressed in the name pattern matching.

This leaves us with **3** unattributed models, that will need to be addressed separately. I will write a issue with regards to the X-series.

LouisLeNezet:
Thanks for your work.
If the Xeon could be added that would be fantastic, as they are the ones used on our HPC.

JosuaCarl:
> Thanks for your work. If the Xeon could be added that would be fantastic, as they are the ones used on our HPC.

They are included, i am just not sure whether they will match with the current pattern matching scheme, because the processors are named: `Xeon E5-2670 v2` in your new TDP file and `Intel® Xeon® Processor E5-2670 v2`. Note the inserted `Processor`in there. I will try to incorporate this into the pattern matching algorithm. The non-ASCII symbols, like `®`are already excluded, but we will have to see how to deal with all edged cases (like including certain words that are not present in some `/proc/cpuinfo` files.

LouisLeNezet:
I will check if this work, and open an issue if necessary 😉

Issue #111, opened 2025-03-11
Add automated testing


JosuaCarl:
The current testing is not executed during pull requests, pushes and so on.

- Testing should happen within a Github workflow

JosuaCarl:
Automated testing already integrated into CI, but fails to run in some cases. Not sure why, but PR #107 is an example where no CI tests were run.

PR #112, opened 2025-03-11
Automated testing within CI

JosuaCarl:
- Added testing after checks to CI workflow
- Used ` jeantessier/test-summary-action@v1` to summarize the results in human readable table
- Bumped versions of `checkout` and `java-setup`
- Added different operating systems into matrix

## Issues
- Addresses #111


JosuaCarl:
I realized, that `gradle check` calls `gradle test`. Therefore this PR is redundant. I will put out a new one addressing the readability of code and results and a separate one bumping the versions.

Issue #101, opened 2024-12-17
Update cpu list regularly


Bastian-Eisenmann:
The provided TDP files do only contain a small bunch of CPUs and there would be a need to update them regular manually.

Rather, there should be some functionality which could be used to update the values in an automated way.
A suitable approach could be to fetch the CPU-Specifications in a regular interval using a script directly from the public available pages of the product developers.

For that purpose may following GitHub project could be used: https://github.com/felixsteinke/cpu-spec-dataset

Bastian-Eisenmann:
AMD-Dataset: https://www.amd.com/en/products/specifications/processors.html (Downloadable as .CSV)
Ampere: https://amperecomputing.com/briefs/ampere-altra-family-product-brief / https://amperecomputing.com/briefs/ampereone-family-product-brief (They have only 2 product series, TDP files needs to be extracted manually or per web-parser)

Intel (little bit more complex because they do not provide a whole dataset as download...):
- https://github.com/toUpperCase78/intel-processors (Last updated 2024, has extracted them from intel product packages manually)
- https://github.com/divinity76/intel-cpu-database/blob/master/README.md (Obsolete, Source unknown)
- https://www.cpubenchmark.net/CPU_mega_page.html (Detailed source unknown)
- https://www.intel.com/content/www/us/en/support/articles/000005505/processors.html (good dataset but desktop only)
- https://www.intel.com/content/www/us/en/support/articles/000028083/processors.html (good dataset but laptop only)
- Product line specific (like intel i7 gen 10) lists can be extracted from the product packages, however this is a huge amount of work but still feasible (round about 20 product pages where the content needs to be manually downloaded). Also it seems like that not all available processors are listed. Entrance: https://www.intel.com/content/www/us/en/products/details/processors.html sample: https://www.intel.com/content/www/us/en/products/details/processors/xeon/e/products.html 
- 

Note: The mentioned scrape from felixsteinke seems currently not to work on intel cpus because they have refactored their page.

LouisLeNezet:
There is also the https://processordb.mit.edu/ that try to summarize all the information from different manufacturer using the database from http://cpudb.stanford.edu/, https://chip-dataset.vercel.app/, https://www.techpowerup.com/cpu-specs/ and https://en.wikichip.org/wiki/WikiChip.
It could be nice to talk with them if we can access their curated dataset if we cite them !

JosuaCarl:
The mentioned databases have some associated problems/challenges: 
[...]

Issue #63, opened 2023-10-17
Restructure code


skrakau:
Some parts of the code became a bit chaotic. Restructure code when main functionality is ready.

JosuaCarl:
Closing issue as it is very general and more specific issues #122 & #151  were opened to address the problems.

Issue #61, opened 2023-10-17
Map CPU model names from LINUX file to the `TDP_CPU` file


Llannelongue:
This is a tricky one and more long term. As it is now, it is likely that the CPU model name returned by LINUX doesn't match exactly the model name in the TDP table. 

One option could be to try and find the standardised model names returned my LINUX and have a mapping table. Or some sophisticated regex to map CPU model names. Open topic!

skrakau:
To start collecting how the differences between the manufacturer marketing name and the `model name` returned in `/proc/cpuinfo` could look like:
- Missing trademark symbols: (R), (TM)
- Additional processor information, such as: XXX 16-Core Processor or CPU @ 3.80GHz
- Suffixes describing different CPU variants, including potentially energy-efficient models, should also be provided by Linux, as far as I understand
- Anything else?

JosuaCarl:
Matching could be done via definition of  distance (in contrast to exact matches). The result would be an estimation with a processor that is the closest in name (although we could define a cutoff value to exclude bad match-ups).

An idea could be to follow these steps:
1. Separate the name of the processor along word separating character such as " ", "-", "_"
2. Ignore certain characters (®/TM)
3. Match separated "words" with "words" from data table
4. Assign a 0.0-1.0 score for similarity (maybe Jaccard similarity + penalty for distance to version number) for each word
5. Normalize to 1

Example:
Real/Manufacturer name: `Intel® Pentium® Gold G7400E Processor` , Provided name: `Intel Pentium Gold G7400` -> Matches to (1.0 + 1.0 + 0.0 + 0.8) / 4 = 0.7  in `Intel® Pentium® G4560`

On could of course then also report the top 5 matched processors in a table if there were no 1.0 match.

skrakau:
In general I wonder, if these are not more systematic differences one could explicitly address. The similarity between `G7400` and `G4560` might not help us that much here. Could we ignore this and just try to explicitly map?  

However, is the above a real example, if yes from where? If the suffix `E` from `G7400E` is missing in the reported `model name` (not in a VM or so), than my assumption from above is wrong.

JosuaCarl:
[...]

PR #104, opened 2025-01-09
Feature/update cpu specific tdp values

Bastian-Eisenmann:
A full automated process seems not to be applicable for updating the cpu specific top files as discussed in: https://github.com/nextflow-io/nf-co2footprint/issues/101

However, in this PR a script and according documentation was added to:
- Download CPU data from AMD, Intel, Ampere or Green Algorithms Project pages
- Merges them together into one .csv file

The currently generated .csv file matches exactly the structure of the Green Algorithms Project. This can later on may be extended to add additional information such as the manufacturer or CPU Thread based TDP information as required in here: https://github.com/nextflow-io/nf-co2footprint/issues/100

PR #118, opened 2025-03-20
Improve handling of CPU info (In memory processing + Pattern matching)

JosuaCarl:
Adds two new general data classes:
  
- Bidirectional Maps (`BiMap`)
  - The goal was to create a mapping Class that incorporated the main functionalities of a standard sorted map, but also works on the premise that the values are (like the keys) a set of unique entries that can be used to access the keys
  - In addition several access methods (like filtering) were defined to interact with the bi-directional map
  - Its intended use is to work as an Index for the `DataMatrix`
  - The main structure behind the `BiMap` are two separate Maps that assign keys-to-values and values-to-keys
  
- Table processing with `DataMatrix`
  - Use the BiMap for indexing the rows and columns in a singular index
  - Uses a list-in-a-list to store the 2D matrix
  - Has access methods to select parts of the `DataMatrix`, get individual values, read in and save the table in CSV format

Adds one specific data class:
- Implementation of `TDPDataMatrix`
  - Accessor methods for larger and complexer TDP table ( enables #101 )
  - Pattern matching with non-ASCII removal (see #61 )

Updated information on `Thermal Design Power (TDP)` of previously untracked processors from the vendor information:
- Added current TDP matrix from https://github.com/qbic-projects/Green_algorithms_assets/tree/main/cpu_info ( #101 )

Incorporated changes into the existing Plugin:
- Addressed pull classes out of Factory mega-class (see #63 )


JosuaCarl:
Handles similar problem to #104

JosuaCarl:
TODO:
- Use latest CPU_Info.csv 
- Pattern matching allowing whitespaces instead of non-ASCII symbols

mapo9 (approved):

nadnein (approved):

plugins/nf-co2footprint/src/main/nextflow/co2footprint/TDPDataMatrix.groovy
```diff
@@ -0,0 +1,212 @@
...
+        return  getTDP(dm, rowID, rowIdx) / getThreads(dm, rowID, rowIdx)
+    }
+
+    /**
+     * Return first name (row index) of the TPDDataMatrix.
+     *
+     * @param dm DataMatrix with TPD values
+     * @return first model name
+     *
+     * @author Josua Carl <josua.carl@uni-tuebingen.de>
```
mapo9, line 205: why do you add the author twice?

mapo9 (approved):
lgtm. only very minor comments

PR #162, opened 2025-04-01
Feature: Added ignoring of 'CPU(s)' during matching

JosuaCarl:
- Adds ignoring of 'CPU(s)' literal during matching


nadnein (approved):
molto bene

PR #163, opened 2025-04-01
Handle: CI & location both explicitly configured

JosuaCarl:
- Refactored the handling of 'ci' and 'location' input
- If 'location' is specified by .config file outside of user-control, 'location = null' can be used to ignore the specified location
- Removed map sanity checks as none were left
- Removed error expectation from test 
- Depends on #159


plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintConfig.groovy
```diff
@@ -36,11 +36,12 @@ class CO2FootprintConfig {
     private String  summaryFile = "co2footprint_summary_${timestamp}.txt"
     private String  reportFile = "co2footprint_report_${timestamp}.html"
     private String  location = null
-    private Double  ci = 475                // CI: carbon intensity
+    private Double  ci = null                // CI: carbon intensity
+    private final Double  default_ci = 475
```
skrakau, line 40: Why does one need an extra variable `default_ci` in this case? Could one not keep it as one variable, and handle everything else when processing the `configMap`? It is a bit confusing, that there are different variables for the defaults in this case, but not for the other cases.
(anyway, it might make sense to adjust this as done for the tdp data and to read this in from the file with the yearly CI values)

JosuaCarl: Yes it would be better to read the CI defaults from the provided table. But that's not done yet. 

I chose to implement it this way because it allows me to check whether the CI value was present in the config. If it were already set to another value I would be unable to get this information from the variable alone.

Another possible implementation would be to check whether the key 'ci' is present in the configMap before iterating over it, or saving the keys separately.

skrakau: OK. Could you add a todo comment to the code then to make clear that this variable will be removed again anyway? then it's probably easier to keep it as it is

skrakau (approved):
Just an additional comment would be nice, but in general LGTM!

PR #169, opened 2025-04-03
CI: Issue templates

JosuaCarl:
- Adds issue templates
- Adds PR templates
- Adds automatic staling of issues
- Adapted from [Nextflow](https://github.com/nextflow-io/nextflow/tree/master/.github)

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 150
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 60
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+# Label to use when marking an issue as stale
+staleLabel: stale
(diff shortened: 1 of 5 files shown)


nadnein (approved):
👍👍

.github/stale.yml
```diff
@@ -0,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 150
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 60
```
mashehu, line 4: Imo it is bad practice to close issues just because they got stale. The underlying problem is then not fixed and it hides it from other users with the same problem. Especially for repos with a low number of maintainers. It is okay, to have issues around if they are problems or feature requests.

JosuaCarl: Upon further inspection maybe the marking as stale could also be dropped, as it contains only a reminder for us that the issue has been around for a long time. Thank you very much for your input.

JosuaCarl: See #174

PR #178, opened 2025-04-16
Feature: Add differentiation of computing machine type

JosuaCarl:
## Related PRs:
- Extends #163 

## Summary
- Introduced new possible key for `.config`: `machineType`
- Introduced reading the `process.executor` to determine `machineType`
- Can be one of `['compute cluster', 'local']`
- Changes PUE according to configuration. Assumes 1.0 PUE if no `machineType` is given.

Checklist
- [x] New functionality is covered by tests
- [x] Documentation of new functionality is added


plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintConfig.groovy
```diff
@@ -31,17 +31,23 @@ import java.util.concurrent.ConcurrentHashMap
...
+    private Double  pue = null              // PUE: power usage effectiveness efficiency, coefficient of the data centre
     private Double  powerdrawMem = 0.3725   // Power draw of memory [W per GB]
     private Boolean ignoreCpuModel = false
-    private Double  powerdrawCpuDefault = 12.0
+    private Double  powerdrawCpuDefault = null
     private String  customCpuTdpFile = null
+    private String  machineType = null      // Type of computer on which the workflow is run ['server', 'local', '']
+
+    // Constants
+    private final Double  default_ci = 475
```
nadnein, line 49: This will become unnecessary in #181

nadnein (approved):
default_ci variable not necessary anymore in #181

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintConfig.groovy
```diff
@@ -85,10 +90,34 @@ class CO2FootprintConfig {
...
             )
         }
+
+        // Keeps ci if already defined, if not uses location if given, fallback to default_ci
         ci ?= location ? retrieveCi(location) : default_ci
 
-        // Reassign default CPU from config
-        if (powerdrawCpuDefault) { cpuData.set(powerdrawCpuDefault, 'default', 'tdp (W)') }
+        // Assign PUE if not already given
+        pue ?= switch (machineType) {
```
skrakau, line 98: There is the Nextflow information `process.executor`, which can be `local` or the cluster or cloud executing system. Could this be used here as well? At least check the `local` case?

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintConfig.groovy
```diff
@@ -31,18 +31,23 @@ import java.util.concurrent.ConcurrentHashMap
...
-    private final Double  default_ci = 475      // TODO: Can be replaced by default in CI fallback table upon addition
-    private Double  pue = 1.67                  // PUE: power usage effectiveness efficiency, coefficient of the data centre
-    private Double  powerdrawMem = 0.3725       // Power draw of memory [W per GB]
+    private Double  ci = null               // CI: carbon intensity
+    private Double  pue = null              // PUE: power usage effectiveness efficiency, coefficient of the data centre
+    private Double  powerdrawMem = 0.3725   // Power draw of memory [W per GB]
     private Boolean ignoreCpuModel = false
     private Double  powerdrawCpuDefault = null
     private String  customCpuTdpFile = null
+    private String  machineType = null      // Type of computer on which the workflow is run ['server', 'local', '']
```
skrakau, line 46: What are the systems we want to differentiate? maybe not `server`, what about `compute cluster`? A server could be part of a cluster or one could use a local server

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintConfig.groovy
```diff
@@ -85,10 +90,34 @@ class CO2FootprintConfig {
...
+        // Keeps ci if already defined, if not uses location if given, fallback to default_ci
         ci ?= location ? retrieveCi(location) : default_ci
 
-        // Reassign default CPU from config
-        if (powerdrawCpuDefault) { cpuData.set(powerdrawCpuDefault, 'default', 'tdp (W)') }
+        // Assign PUE if not already given
+        pue ?= switch (machineType) {
+            case 'local' ->  1.0
+            case 'server' -> 1.67
+            default -> 1.67
```
skrakau, line 101: Would it not be more reasonable to use `local` as the default case?

skrakau (changes requested):
Looks good, but I have a few comments and the documentation about this new parameter is still missing.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintConfig.groovy
JosuaCarl: This is true, I am unsure how to handle this. I would assume that the nomenclature with 'server' and 'cluster' could get across what is meant. But may be 'server cluster' or 'computing cluster' is the most unambiguous.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintConfig.groovy
JosuaCarl: I did not know that, I'll check.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintConfig.groovy
skrakau: What's wrong with "compute cluster"? In any case, to me something with computing seems more suitable than "server cluster"

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintConfig.groovy
JosuaCarl: I think `Nextflow` does it and so should we. I was just hesitant about changing it.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintConfig.groovy
```diff
@@ -85,10 +90,34 @@ class CO2FootprintConfig {
...
+        }
+
+        // Reassign values based on machineType
+        if (machineType) {
+            if (supportedMachineTypes.contains(machineType)) {
+                cpuData.fallbackModel = "default $machineType"
+            }
+            else {
+                log.warn(
```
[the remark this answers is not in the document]
JosuaCarl: No, I don't think so. I think this is a general problem with fallback values (see the issue #188) I opened up to address this. I would reimplement all fallback warnings as errors, if a parameter `useFallback` is set to `false` (defaults to `true`)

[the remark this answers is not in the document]
JosuaCarl: Maybe in this case an error would be better because nothing is matches. This could change however, when using the `process.executor` parameter from Nextflow, because the names would need to be matched against their typical target.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintConfig.groovy
JosuaCarl: I think this would be a good approach, but would implement the matching at some other point and raise this as a feature issue for now because it seems a bit out of scope of this PR to implement a matcher for the possible values of `process.executor`

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintConfig.groovy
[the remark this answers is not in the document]
JosuaCarl: Error is now thrown.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintConfig.groovy
JosuaCarl: I decided to implement it with this PR in the form of a matcher `matchProcessExecutor` function in the config. I am not sure whether this should be in the Config, but could not think of a better place. The standard process.executor is empty, despite 'local' being named as the default in the Nextflow documentation. I implemented it such that empty values are no problem and fall back to our default.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintConfig.groovy
JosuaCarl: Used compute cluster.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintConfig.groovy
```diff
@@ -83,10 +102,37 @@ class CO2FootprintConfig {
...
+
+        // Reassign values based on machineType
+        if (machineType) {
+            if (supportedMachineTypes.contains(machineType)) {
+                cpuData.fallbackModel = "default $machineType"
+            }
+            else {
+                String message =
+                        "machineType '${machineType}' is not supported. Please chose one of ${supportedMachineTypes}." +
+                        "Using fallbacks: pue=${pue} & fallbackModel=${cpuData.fallbackModel}."
```
skrakau, line 127: Fallback values should not be used in case of an error

docs/usage/parameters.md
```diff
@@ -37,3 +37,6 @@ Default: `false`.
 - `powerdrawCpuDefault`: the default value used as the power draw from a computing core.
 This is only applied if the parameter `ignoreCpuModel` is set or if the retrieved `cpu_model` could not be found in the given CPU TDP data.
 Default: 12.0.
+- `machineType`: the type of machine the computation is executed upon. Has to be in ['compute cluster', 'local', ''].
+  `process.executor` is used to infer the `machineType` when no value is set.
```
skrakau, line 41: Maybe describing what this is could help? For example

```suggestion
  The Nextflow config option `process.executor` is used to infer the `machineType` when no value is set.
```

skrakau (approved):
Great! Just one small suggestion

PR #181, opened 2025-04-22
Feature: Real-time Carbon intensity values from Electricity Maps API + Fallback

nadnein:
## Summary: 
This pull request improves carbon intensity (CI) handling in the CO₂ footprint calculation by introducing a new fallback table for CI values and integrating real-time carbon intensity data from the Electricity Maps API.

## Features: 
- [x]  implement CI value handling: CI specified in config > if API key provided use real-time CI > if location provided use CI from table for location > else use global CI from fallback table 
- [x] use [electricity maps API](https://portal.electricitymaps.com/dashboard)
- [x] get CI per task 
- [x] add per task CI to trace file 
- [x] store API key as nextflow secrets 
- [x]  add per task CI to html report table
- [ ] add graph of CI over time to summary file (?)

## Relates to:  
- Closes #148 
- Approach #96 of @tillenglert was further implemented

PR #201, opened 2025-05-16
Fix: Fixed HTML report rendering and improve time formatting

nadnein:
## Summary:
This PR fixes issues in the HTML report generation and improves time unit formatting for better readability and consistency.

## Changes:
- **Fixed tree month value rendering:**  
  The tree_month field in the HTML report was sometimes empty. This has been resolved.
- **Improved CI value display:**  
  When real-time CI data is used, the field previously showed a non-informative closure string (e.g. nextflow.co2footprint.CIValueComputer$_computeCI_closure1@...). It now shows "dynamic".
- **Time unit formatting:**  
  - Tree month values are now consistently displayed in months, instead of falling back to seconds or other units.
  - Fixed `toReadableTimeUnits()` function so that it no longer returns an empty string in some cases. Refactored it to make it easier to understand.


plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2EquivalencesRecord.groovy
```diff
@@ -17,9 +17,7 @@ class CO2EquivalencesRecord {
     String getCarKilometersReadable() { Converter.toScientificNotation(carKilometers) }
 
     Double getTreeMonths() { treeMonths }
-    String getTreeMonthsReadable() {
-        Converter.toReadableTimeUnits(treeMonths, 'months', 's', 'years', 1, 3)
-    }
+    String getTreeMonthsReadable() { Converter.toScientificNotation(treeMonths) }
```
JosuaCarl, line 20: I would like the `toReadableTimeUnits` to be used here to get the final printed result. A bug in the method should not result in exclusion.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/FileCreators/CO2FootprintReport.groovy
JosuaCarl, line 1: Could be shortened a bit. I find it more readable, but is not a must.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/Converter.groovy
```diff
@@ -96,59 +96,57 @@ class Converter {
...
-                    value, unit,
-                    smallestUnit, units[units.indexOf(largestUnit) - 1],
-                    threshold, numSteps, readableString
-            )
+            String result = readableString.trim()
+            return result ? result : "0${smallestUnit}"
         }
+
+        // Otherwise, continue with the next smaller unit
+        String nextLargestUnit = units[largestIdx - 1]
```
JosuaCarl, line 145: :+1:

plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/Converter.groovy
```diff
@@ -96,59 +96,57 @@ class Converter {
...
-            return readableString.trim()
-        }
-        else {
-            return toReadableTimeUnits(
-                    value, unit,
-                    smallestUnit, units[units.indexOf(largestUnit) - 1],
-                    threshold, numSteps, readableString
-            )
+            String result = readableString.trim()
+            return result ? result : "0${smallestUnit}"
```
JosuaCarl, line 141:
```suggestion
           return readableString ? readableString.trim() : "0${smallestUnit}"
```
This does not work, if single white spaces are given to readableString, but that shouldn't happen. Does that happen?

plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/Converter.groovy
```diff
@@ -96,59 +96,57 @@ class Converter {
...
-     * Converts the given unit to a readable time string with options to limit what is shown.
-     *
-     * @param value The time as a number in original given unit
-     * @param unit Given unit of time
-     * @param smallestUnit Smallest desired unit to be reported
-     * @param largestUnit Largest desired unit to be reported
-     * @param threshold Smallest value to be reported (remainders are passed on to smaller units)
-     * @param maximumSteps Maximum number of valid time steps to be reported
-     * @return String of the readable time
-     */
```
JosuaCarl, line 109: Why delete the docstring?

plugins/nf-co2footprint/src/main/nextflow/co2footprint/FileCreators/CO2FootprintReport.groovy
```diff
@@ -157,8 +157,13 @@ class CO2FootprintReport extends CO2FootprintFile{
...
 
         // Render JSON
-        List<String> options = all_options.collect {name, value ->
-            "{ \"option\":\"${name}\", \"value\":\"${value as String}\" }" as String
+        List<String> options = all_options.collect { name, value ->
+            def valueStr = value as String
+            // If value is a Closure (e.g. in case API calls are being made for each task), replace with 'dynamic'
+            if (value instanceof Closure) {
+                valueStr = 'dynamic'
+            }
```
JosuaCarl, line 165:
```suggestion
            // If value is a Closure (e.g. in case API calls are being made for each task), replace with 'dynamic'
            String valueStr = value instanceof Closure ? 'dynamic' : value as String
```

plugins/nf-co2footprint/src/test/nextflow/co2footprint/CO2FootprintObserverTest.groovy
```diff
@@ -203,7 +203,7 @@ class CO2FootprintObserverTest extends Specification{
                 'cpu_model', 'cpu_usage', 'requested_memory'
         ]
         traceLines[1].split('\t') as List<String> == [
-                '111', 'null', 'null', '14.61  Wh', '6.94  g', '1.0ms', '1', '12.0', 'Unknown model', '100.0', '7.0 B'
+                '111', 'null', 'null', '14.61  Wh', '6.94  g', '1ms', '1', '12.0', 'Unknown model', '100.0', '7.0 B'
```
JosuaCarl, line 206: Why did this change? I see no reason in the code why this should be different.

plugins/nf-co2footprint/src/test/nextflow/co2footprint/utils/ConverterTest.groovy
```diff
@@ -33,9 +33,9 @@ class ConverterTest extends Specification  {
         where:
         value   || unit     || smallestUnit     || largestUnit  || smallestValue    || maximumSteps     || expected
         1.0     || 'days'   ||  'h'             ||  'days'      || 0.0              || null             || '1day'
-        2.1     || 'days'   ||  'min'           ||  'days'      || null             || null             || '2days 2h 24.000min'
+        2.1     || 'days'   ||  'min'           ||  'days'      || null             || null             || '2days 2h 24min'
         3600.0  || 's'      ||  's'             ||  'min'       || null             || 1                || '60min'
-        3602.1  || 's'      ||  's'             ||  'min'       || null             || null             || '60min 2.1000s'
-        3602.1  || 's'      ||  's'             ||  'min'       || null             || 1                || '60.035min'
+        3602.1  || 's'      ||  's'             ||  'min'       || null             || null             || '60min 2s'
+        3602.1  || 's'      ||  's'             ||  'min'       || null             || 1                || '60min'
```
JosuaCarl, line 39: Same thing, this should not change. Why did it?

JosuaCarl (changes requested):

plugins/nf-co2footprint/src/test/nextflow/co2footprint/utils/ConverterTest.groovy
nadnein: Why should there be trailing zeros? E.g in 2.1000s?

plugins/nf-co2footprint/src/test/nextflow/co2footprint/CO2FootprintObserverTest.groovy
nadnein: Why should there be training zeros? I thought its more convenient to print this then as 1ms. But I can also revert the changes I didn't wanted to change your function if you don't like the changes

plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/Converter.groovy
nadnein: Was not on purpose oops

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2EquivalencesRecord.groovy
JosuaCarl: Because the reader of the document will not get what 0.03 months are, but 1 day is readable.

plugins/nf-co2footprint/src/test/nextflow/co2footprint/utils/ConverterTest.groovy
JosuaCarl: Agreed, trailing 0s could be deleted, but not `60.035min` ->  `60min`

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2EquivalencesRecord.groovy
nadnein: but then it says "1 day tree month" in the report :D That's what I meant

plugins/nf-co2footprint/src/test/nextflow/co2footprint/CO2FootprintObserverTest.groovy
nadnein: Example for targetValue 1.6

Screenshot 2025-05-19 at 10 24 06

plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/Converter.groovy
nadnein: Yes can happen, therefore the code return "0${smallestUnit}" as a fallback.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/Converter.groovy
```diff
@@ -96,59 +96,57 @@ class Converter {
...
+        int smallestIdx = units.indexOf(smallestUnit)
+        int largestIdx = units.indexOf(largestUnit)
+        numSteps = (numSteps == null) ? (largestIdx - smallestIdx) : (numSteps - 1)
 
-        // Calculate the time in the target unit
+        // Convert value to the current target unit
         String targetUnit = largestUnit
         BigDecimal targetValue = convertTime(value as BigDecimal, unit, targetUnit)
-        def targetValueFormatted = Math.floor(targetValue) as Integer
+        int targetValueInt = Math.floor(targetValue) as Integer
```
JosuaCarl, line 119: Ah, here is the code that makes the difference in the output:
```suggestion
        def targetValueInt = Math.floor(targetValue) as Integer
```

plugins/nf-co2footprint/src/test/nextflow/co2footprint/utils/ConverterTest.groovy
nadnein: ok 👍 I will change it back

JosuaCarl (changes requested):

plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/Converter.groovy
JosuaCarl: But is `'  '` returned or `''`. The latter one I can see, but not the former.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2EquivalencesRecord.groovy
JosuaCarl: Does it ? Then I would change the wording in the report.

JosuaCarl: You're right. It is still present here:
https://github.com/nextflow-io/nf-co2footprint/blob/cfc4421b909e96f7df9bfcaab4005221b67186c6/plugins/nf-co2footprint/src/resources/CO2FootprintReportTemplate.html#L203

nadnein: Ok maybe we have to agree on that with Sabrina because "Tree Month" is a defined equivalence statistic (at least they write it like this in Green Algorithms). @skrakau

plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/Converter.groovy
nadnein: I think the latter yes

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2EquivalencesRecord.groovy
nadnein: OK then I will also change it to this format in the HTML report.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/Converter.groovy
```diff
@@ -116,23 +128,25 @@ class Converter {
...
         // Only add to output if above threshold or no threshold set
-        if (threshold == null || targetValueInt > threshold) {
-            value = targetValue - targetValueInt
+        if (threshold == null || targetValueFormatted > threshold) {
+            value = targetValue - targetValueFormatted
             unit = largestUnit
-            readableString += " ${targetValueInt}${targetUnit}"
+            // Format to 2 decimals, remove trailing zeros
+            String formattedValue = new BigDecimal(targetValueFormatted.toString()).setScale(2, BigDecimal.ROUND_HALF_UP).stripTrailingZeros().toPlainString()
+            readableString += readableString ? " ${formattedValue}${targetUnit}" : "${formattedValue}${targetUnit}"
```
JosuaCarl, line 149:
```suggestion
            readableString += "${formattedValue}${targetUnit}"
```
Is there something I'm not seeing. This is equal to:
```groovy
if(readableString {
  readableString += " ${formattedValue}${targetUnit}"
}
else {
  readableString += " ${formattedValue}${targetUnit}"
}
```
right ? Which results both times in the same output.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintConfig.groovy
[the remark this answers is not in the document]
nadnein: Yes in #181, I just had to change that here so that the code runs with my config

plugins/nf-co2footprint/src/main/nextflow/co2footprint/FileCreators/CO2FootprintReport.groovy
```diff
@@ -158,12 +158,8 @@ class CO2FootprintReport extends CO2FootprintFile{
 
         // Render JSON
         List<String> options = all_options.collect { name, value ->
-            def valueStr = value as String
-            // If value is a Closure (e.g. in case API calls are being made for each task), replace with 'dynamic'
-            if (value instanceof Closure) {
-                valueStr = 'dynamic'
-            }
-            "{ \"option\":\"${name}\", \"value\":\"${valueStr}\" }"
+            def valueStr = (value instanceof Closure) ? 'dynamic' : value as String
```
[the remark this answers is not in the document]
nadnein: `value` has type Object before

plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/Converter.groovy
nadnein: Once there is a leading space, once there is not

plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/Converter.groovy
nadnein: I changed it

plugins/nf-co2footprint/src/main/nextflow/co2footprint/FileCreators/CO2FootprintReport.groovy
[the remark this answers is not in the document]
JosuaCarl: But in the end we always get a String for the `valueStr` variable. Either from the `value` Object converted to String, or from 'dynamic'.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/Converter.groovy
JosuaCarl: Ah, I see. I think I missed it because my screen size resulted in a line break at the position. :+1:

plugins/nf-co2footprint/src/main/nextflow/co2footprint/FileCreators/CO2FootprintReport.groovy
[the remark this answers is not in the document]
nadnein: Ah yes sorry I had misread your correction

JosuaCarl (approved):
:+1:

JosuaCarl (approved):

Issue #197, opened 2025-05-14
Config bugs in CO2FootprintConfig and TDP data after changes to naming


JosuaCarl:
- When passing `''` to machineType, machineType is still overwritten
  `machineType ?= ` should change to `if(machineType == null) {} else {}`
   or should we just drop `''` as a possible input ?
- TDP data table still contains `default server` instead of `default compute cluster`

PR #204, opened 2025-05-19
Chore: Renamed server to 'compute cluster'

JosuaCarl:
### Related Issues
- Addresses #197 

### Summary
Renamed all instances of 'server' to 'compute cluster' in TDP table


JosuaCarl:
Keep in mind that this should be changed over at https://github.com/qbic-projects/Processor_specifications_collector/pull/18

PR #206, opened 2025-05-19
Fix: Skip comparison to executor, when `''` is explicitly specified as `machineType`

### LINKED ISSUE #197: Config bugs in CO2FootprintConfig and TDP data after changes to naming

JosuaCarl:
## Related Issues
- Closes #197 

## 🎯 Motivation
- Intended behavior
- Give user the possibility to skip the comparison to Nextflows's executor

## 📋 Summary of changes
- One line change to not replace `''`, but only `null` as a `machineType`

## ✅ Checklist
- [x] Documentation reflects behaviour


nadnein:
I thought we wanted to completely discard the option to set `''` as a value for the machineType in the user config because it's redundant to make use of the `pue` config option. And I find it also confusing to have `''` as a value option.

JosuaCarl:
I oriented it at the documentation. I thought we did say that it could be useful to circumvent the `trace.executor` checking and then going to the default value. I thought we agreed on this because of the documentation you merged:
https://github.com/nextflow-io/nf-co2footprint/blob/0d48fa7c2bd343d592f2c6b607a33e803b3e65b9/docs/usage/parameters.md?plain=1#L63-L71

To be honest, I would just leave it in, as just setting the PUE is always an option that overwrites the others. We could change both (docs & behaviour) at some other point.

nadnein:
Yes but I thought that we agreed there that the `''` option is redundant and that you write an issue to change it soon. But then I will approve the PR now but write an issue :)

nadnein (approved):

PR #210, opened 2025-05-20
Refactor: Moving call to get total Memory size 

JosuaCarl:
## 🎯 Motivation
- Avoid execution of calls that are mostly unnecessary

## 📋 Summary of changes
- Moved OS call into the case that memory is exceeded

## ✅ Checklist


nadnein (approved):
👍

PR #215, opened 2025-05-21
Feature: Break on fallback to default

### LINKED ISSUE #188: [Feature]: Give the user the possibility to break the program when using default values
### Describe the feature

Create a possibility for the user to catch non matching CPU names or locations early on.

### Use case(s):

This avoids running long processes and discovering in hindsight that the default TDP was used, potientially yielding inexact results.

### Suggested implementation:

Implement a `useFallback=true` parameter in the config to have the calculation break if it does not find the CPU model / location for the CI. The `fallback` option has to be implemented into the matching of CI and TDP.

JosuaCarl:
## Related Issues
- Closes #188 

## 🎯 Motivation
- Give the user the chance to avoid unnecessary computations, when they don't want to accept a default value.

## 📋 Summary of changes
- Added the `fallbackToDefault` config parameter to define whether the default value should be used as a fallback when no match is found.

## ✅ Checklist
- [ ] Added `falllbackToDefault` to `CI`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Speedier Report generation through refactoring of Co2 aggregation
 - Merge provided custom TDP matrices into the old `TDPDataMatrix`, instead of fully replacing it
 - Added metric to indicate newly generated / non-cached CO2 emissions into report
+- Added support for breaking the program when falling back to defaults
 
 ## Bug Fixes:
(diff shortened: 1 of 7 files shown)


JosuaCarl:
Abandoned because no clear reason to implement. Also: Which fallback would be ok and which wouldn't.

PR #225, opened 2025-06-03
Add null checks, fallbacks, and logging to CO₂ calculation 

### LINKED ISSUE #219: [Bug]: Warning not printed when requestedMemory is null
This warning is not printed when `requestedMemory` is `null`:

https://github.com/nextflow-io/nf-co2footprint/blob/f8a8180e07ca887826efbbf45ecfb9659da2b7c3/plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintComputer.groovy#L94-L98

nadnein:
This PR enhances the CO₂ footprint computation in the `CO2FootprintComputer` class with null checks, fallbacks, and logging.

## Main changes
- Situations where trace field values that are important for CO2 footprint calculation are null are now handled by using in this case a default value and logging a warning.
- Improved fallback logic for memory assignment:  If the requested memory is missing or the peak memory usage exceeds the requested memory, the code assigns the available system memory as the requested memory and logs a warning. If the available system memory can't be determined or is zero, it throws an exception and logs an error.
- Add tests for memory assignment and fallback behavior.

## Related Issues
- Closes #219 

## ✅ Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ## Bug Fixes:
 - Adjusted rendering of flights to deliver percentage < 1.0 flights and number of flights afterwards
+- Add null checks, fallbacks, and logging to CO₂ calculation in `CO2FootprintComputer` class 
 
 ## Misc
 - Moved call to OS when memory is exceeded into this case to avoid unnecessary calls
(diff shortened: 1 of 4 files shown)


JosuaCarl (approved):
Good changes, good test coverage, better comments :+1:

PR #228, opened 2025-06-03
Delete COPYING file as it was redundant to LICENSE file in main repository

nadnein:
The COPYING file with the Apache License for the source code is deleted as it is redundant to LICENSE file of the source code. It also appeared therefore two times in the header on the GitHub repository (see screenshot).

### Screenshot:
Screenshot 2025-06-03 at 15 27 53


JosuaCarl (approved):
:+1:

PR #227, opened 2025-06-03
Update `collectIndices` method to address bug 

nadnein:
The updated `collectIndices` method in `DataMatrix` now robustly matches keys by comparing their string representations, ensuring that keys with equivalent string values (even if of different types) are correctly identified in the BiMap. If a key is not found, it logs a clear warning and skips that key, preventing downstream errors caused by missing indices. 

## Related Issues
- Closes #226, as this was the root cause for the bug 

## ✅ Checklist
- [x] New functionalities are covered by tests
- [x] Class structure in `test` reflects class structure in `main`
- [x] Ensure all tests pass (`.\gradlew check`)

PR #230, opened 2025-06-04
Fix unintended assignment to GString class 

nadnein:
The unintended assignment of the GString class to the `TDPDataMatrix.fallbackModel` variable was changed to the String class, which caused the bug described in #227. Therefore, the root cause of this bug was the assignment of the GString class instead of the String class, and not the `collectIndices` method itself.

## Summary:
- Change GString assignment to String 
- Add test that ensures correct fallback row is used for unknown CPU models in TDPDataMatrix

## Related issues:  
- Closes #227 

## ✅ Checklist


JosuaCarl (approved):
Long live the `Jstring` 🥇

PR #234, opened 2025-06-06
Fix: removed '' machineType input option

### LINKED ISSUE #208: [Refactoring]: '' as a value option for the machineType parameter in the plugin config redundant
The option to set `''` as a value for the `machineType` in the user config is redundant to making use of the `pue` config option, and could be removed.

For orientation, the current usage of the `machineType` param is described in the docs like this: 
https://github.com/nextflow-io/nf-co2footprint/blob/0d48fa7c2bd343d592f2c6b607a33e803b3e65b9/docs/usage/parameters.md?plain=1#L63-L70

And for `pue` like this: 
https://github.com/nextflow-io/nf-co2footprint/blob/0d48fa7c2bd343d592f2c6b607a33e803b3e65b9/docs/usage/parameters.md?plain=1#L34-L36

nadnein:
In this PR, all files that had to be changed for removing `''`as input option for the `machineType` were updated. 

## Related Issues
- Closes #208 

## ✅ Checklist

### DIFF
--- a/docs/usage/configuration.md
+++ b/docs/usage/configuration.md
@@ -23,11 +23,11 @@ co2footprint {
   traceFile = "${params.outdir}/pipeline_info/co2footprint_trace_${co2_timestamp}.txt"
   summaryFile = "${params.outdir}/pipeline_info/co2footprint_summary_${co2_timestamp}.txt"
   reportFile = "${params.outdir}/pipeline_info/co2footprint_report_${co2_timestamp}.html"
-  location = '<your_zone_code>'     // replace with your zone code
-  ci = <your_ci>                    // replace with carbon intensity (gCO2eq/kWh)
-  apiKey = secrets.EM_API_KEY       // set your API key as Nextflow secret with the name 'EM_API_KEY'
-  pue = <your_pue>                  // replace with PUE of your data center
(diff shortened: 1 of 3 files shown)


nadnein:
What would you change in the code? I thought there was nothing that had to be changed. The only thing we could do is to verify that the value set for `machineType` is one of the "allowed" ones, but we also didn't do that yet for the other config params.

JosuaCarl:
See here:
https://github.com/nextflow-io/nf-co2footprint/blob/092585b08b9795c62d1ce0380ba48df20fddecc0/plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintConfig.groovy#L46-L48
and here:
https://github.com/nextflow-io/nf-co2footprint/blob/092585b08b9795c62d1ce0380ba48df20fddecc0/plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintConfig.groovy#L101-L111

The whole block does not get executed if `machineType` is an empty String `''`. So in this case it would change nothing, but I would at least take the option out of the `supportedMachineType`s options.

JosuaCarl (approved):
:+1:

PR #229, opened 2025-06-03
Conversion method assets fix

JosuaCarl:
## Related Issues
- Closes #199 

## 🎯 Motivation
- Align Javascript with Groovy code (Converters)
- Fix IDE warnings in HTML template
- Make Webstack code more readble

## 📋 Summary of changes
- Added lots of comments
- Moved styling CSS to own file
- Reworked the Javascript number converters to mirror the Groovy code (mainly `toReadableTimeUnits()`)

## ✅ Checklist
- [x] New functionalities are covered by tests
- [x] Class structure in `test` reflects class structure in `main`
- [x] `CHANGELOG.md` is updated with a note on your changes
- [x] Ensure all tests pass (`.\gradlew check`)

PR #237, opened 2025-06-11
Feature: Incorporating newly generated emissions (vs. cached) into the report

### LINKED ISSUE #62: Inclusion of the carbon footprint of cached processes
This is to discuss how to present the carbon footprint of cached processes.

Simple example: 3 processes [P1] [P2] [P3]
- First run of the pipeline, all three are run
- Second run of the pipeline, cached [P1] is used, [P2] and [P3] are run

When presenting the carbon footprint of run (2), we can either:
- Take the carbon footprint of [P1] for run (1) and add the carbon footprints of [P2] and [P3] from run (2).
- Or only add the carbon footprints of [P2] and [P3] and ignore the impact of [P1].

Option 1 gives a better estimate of the total carbon footprint of the pipeline if we were to run it again from start to finish on new data let's say. But option 2 gives a more accurate estimate of the true carbon footprint of running step (2). And if adding run (1) + run (2), option 2 should be used (otherwise the footprint of [P1] would get double counted even though it was only run once). 

It seems to depend a lot on what users want to do with this information, so perhaps best to give both information in the report so that users can decide what to do?

JosuaCarl:
## TODO
- Wait for #229 and rebase

## Related Issues
- Closes #62

## Related Pull-requests
- Builds upon #229

## 🎯 Motivation
- Capture the cached processes separately for the report file.  Previously, cached processes were treated like normal. This could cause some double counting

## 📋 Summary of changes
- Modified aggregator to include `TraceRecord` and thus enabling the tracing of CO2 Records from CACHED processes
- Added an explicit line to the summary, stating that the CACHED values are included
- Added newly generated / non-cached emissions to plot and general metric cards

## Screenshots
Screenshot from 2025-06-17 16-56-31
Screenshot from 2025-06-17 16-56-49
Screenshot from 2025-06-17 16-57-05
Screenshot from 2025-06-17 16-58-37

## ✅ Checklist


nadnein:
Can you add a Screenshot? :))

nadnein:
I think it's a nice addition! It would perhaps be clearer and more compact if you could fold out the stats for ‘without cached tasks’ so that they are not directly there, but you already said that this is not so easy.
And I would name the boxplots for the processes without cached processes `without cached` and not `not cached`

JosuaCarl:
The plot is the same with the addition of `without cache` values. The new plot just hast one value, that's why there is no box. But it works with boxes as well.

I'm unsure what you meant by the 2nd comment. The reason why we included this is to avoid people stopping a run, then resuming it and then adding the two total emissions to determine the total.

skrakau:
OK I see, I would not mix `total` and `without cached` within on plot. For larger workflows, this would get quite confusing I think. You could add different tabs (as it was done at the beginning for CO2 emissions and energy consumption https://github.com/nextflow-io/nf-co2footprint/pull/44, and as it is done for the Nextflow execution report), one for `total` and one for `without cached`. What do you think?

nadnein:
Looks good! The only thing I'd suggest is adding a note to the title of the 'without cached tasks' plot - something like 'CO₂ emissions & energy consumption (without cached tasks)' - so it's clearly distinguishable from the other plot and doesn't look identical.

PR #252, opened 2025-07-02
Add color to logging messages

nadnein:
# PR: Improve Log Output Readability and Enable Colored Logging

This PR improves the readability of log output by adding a `logback.xml` configuration file to enable colored log levels in the console.

### Screenshot: 
Screenshot 2025-07-02 at 14 25 13


JosuaCarl (approved):
:+1:

Small comment: If we overwrite the Nextflow Logger with this and we get reports about it being a problem, we would need to find another solution / take it back

PR #256, opened 2025-07-07
Add 'default cloud' row to tdp table

nadnein:
## Summary:
This PR adds a "default cloud" row to the TDP table, duplicating the entries from the "default compute cluster" row.
This change ensures that when "cloud" is assigned as the `machineType`, the program can find the corresponding "default cloud" row in the TDP table and avoids errors during lookup.


JosuaCarl (approved):
:+1:

PR #269, opened 2025-07-21
Fix: Error logging on providing `customTDPTable`

JosuaCarl:
## Related Issues
- Error messages on customTDPTable loading

## 🎯 Motivation
- Avoid multiple errors being thrown when reading in an incomplete TDP table

## 📋 Summary of changes
- Avoided try-catch in favor of beforehand index comparison
- Improved logging to depict ID that was not matched on error

## ✅ Checklist

### DIFF
--- a/plugins/nf-co2footprint/src/main/nextflow/co2footprint/DataContainers/DataMatrix.groovy
+++ b/plugins/nf-co2footprint/src/main/nextflow/co2footprint/DataContainers/DataMatrix.groovy
@@ -272,40 +272,45 @@ class DataMatrix implements Matrix {
         return this.columnIndex.valueSet().sort().collect {i -> this.columnIndex.getKey(i)}
     }
 
+    Integer getIdx(Object id, String orientation, boolean checkOccurrence=true) {
+        BiMap<Object, Integer> index = switch (orientation) {
+            case 'row' -> this.rowIndex
+            case 'column' -> this.columnIndex
+            default -> null
+        }
+
+        Integer idx = index?.getValue(id)
+
+        if (checkOccurrence && idx == null) {
+            final String message = "${orientation.capitalize()} ID `${id}` not found in the ${orientation}-index `${index}`."
+            log.error(message)
+            throw new IllegalArgumentException(message)
+        }
+
+        return  idx
+    }
+
     /**
      * Get a value from the matrix by row and column.
      * Row and column can be names or integer indices, controlled by rowRawIndex/columnRawIndex.
      */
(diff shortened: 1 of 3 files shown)


plugins/nf-co2footprint/src/main/nextflow/co2footprint/DataContainers/DataMatrix.groovy
```diff
@@ -272,40 +272,45 @@ class DataMatrix implements Matrix {
...
+            default -> null
         }
 
-        // Resolve column index
-        column = columnRawIndex ? column as Integer : this.columnIndex.getValue(column)
-        if (column == null) {
-            final String message = "Column '${columnRawIndex ? column : column as String}' not found in the column index."
+        Integer idx = index?.getValue(id)
+
+        if (checkOccurrence && idx == null) {
```
nadnein, line 284: Why do we need a `checkOccurrence` variable? Shouldn't the error always be thrown in that case?

plugins/nf-co2footprint/src/main/nextflow/co2footprint/DataContainers/DataMatrix.groovy
```diff
@@ -272,40 +272,45 @@ class DataMatrix implements Matrix {
...
-     * Row and column can be names or integer indices, controlled by rowRawIndex/columnRawIndex.
-     */
-    Object get(Object row, Object column, boolean rowRawIndex=false, boolean columnRawIndex=false) {
-        // Resolve row index
-        row = rowRawIndex ? row as Integer : this.rowIndex.getValue(row)
-        if (row == null) {
-            final String message = "Row '${rowRawIndex ? row : row as String}' not found in the row index."
-            log.error(message)
-            throw new IllegalArgumentException(message)
+    Integer getIdx(Object id, String orientation, boolean checkOccurrence=true) {
```
nadnein, line 275: Can you update the docstring? It's hard to get easily what the method does. And maybe rename `id`to `rowOrColumnId`?

nadnein (changes requested):

plugins/nf-co2footprint/src/main/nextflow/co2footprint/DataContainers/DataMatrix.groovy
JosuaCarl: Not when we would expect that there could be no match. For example in `putRow` this is the case.

nadnein (approved):

PR #273, opened 2025-07-28
Fix: downgrade logging

nadnein:
## Summary:

- Updated the logic in CO2FootprintConfig to silently skip the 'params' key when assigning values from the config map
- When a trace field value is missing, a warning is now logged only the first time it occurs. For subsequent missing values of the same field, only a debug-level message is logged (so they appear in the nextflow log file) 
- Updated the `decide` function in `DeduplicateMarkerFilter` class: Now supports deduplication based on a custom dedupKey passed via log parameters. Only logs the first occurrence of a message with the same dedupKey; subsequent duplicates are logged at TRACE level as [DUPLICATE]. Allows optional custom trace messages via log parameters.
- added test


plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintComputer.groovy
```diff
@@ -172,7 +187,7 @@ class CO2FootprintComputer {
      * @return CO2EquivalencesRecord with estimations for sensible comparisons
      */
     static CO2EquivalencesRecord computeCO2footprintEquivalences(Double totalCO2) {
-        final BigDecimal gCO2 = totalCO2 as BigDecimal / 1000 as BigDecimal       // Conversion to [g] CO2
+        final BigDecimal gCO2 = totalCO2 as BigDecimal / 1000 as BigDecimal // Convert to gCO2
```
JosuaCarl, line 190:
```suggestion
        final BigDecimal gCO2 = totalCO2 as BigDecimal / 1000 as BigDecimal // Convert to [gCO2]
```
[] marks a unit. It's not a big deal, but I find it easier to find/read when units are marked.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintFactory.groovy
```diff
@@ -98,6 +98,7 @@ class CO2FootprintFactory implements TraceObserverFactory {
         // Read the plugin version
         setPluginVersion()
         log.info("nf-co2footprint plugin  ~  version ${this.pluginVersion}")
+        log.info('🔕 Repeated logging messages for different tasks are shown only once in the console; further occurrences are written to the `.nextflow.log` file.')
```
JosuaCarl, line 101: :+1:

JosuaCarl: Although, this only describes standard behaviour. If the console appender is set to DEBUG level or TRACE level it will occur in the console. We could leave it like it is, or write something like:
"Reapeated logging messages are logged one time at given level (displayed in the console) and then deescalated to DEBUG level (only written to `nextflow.log` file)"

But I find my phrasing clunky and maybe we should also keep it as is.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/Logging/DeduplicateMarkerFilter.groovy
```diff
@@ -15,6 +16,7 @@ import java.util.concurrent.atomic.AtomicInteger
  * Motivation: Removes duplicates in some warnings, to avoid cluttering the output with repeated information.
  * Example: If the CPU model is not found it should only be warned once, that a fallback value is used.
  */
+@Slf4j
```
JosuaCarl, line 19:
```suggestion
```
This should not be there. We would not want an endless loop in the Turbo Filter by logging from it.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/Logging/DeduplicateMarkerFilter.groovy
```diff
@@ -69,31 +71,35 @@ class DeduplicateMarkerFilter extends TurboFilter {
...
             int currentOccurrences = occurrences.incrementAndGet()
 
+            // Allow up to allowedOccurrences, then log further as TRACE
             if (currentOccurrences <= allowedOccurrences) {
                 return FilterReply.ACCEPT
             }
-            // Send a TRACE message when the message was not accepted
-            logger.trace('[DUPLICATE] ' + format, params)
+
+            logger.trace('[DUPLICATE] ' + logMessage, params)
```
JosuaCarl, line 100: You wanted to deescalate to DEBUG level, right?

JosuaCarl (commented):
Minor comments

PR #283, opened 2025-08-05
Fix: round numbers and remove bold from warnings

nadnein:
## Summary: 
- Round numbers in log statements 
- Remove bold from warnings —> otherwise removes format from Slf4j logging  —> yellow colored font becomes white after bold text


JosuaCarl (approved):
:+1: Minor comments.

PR #284, opened 2025-08-06
Fix: Ensure consistency between reports

nadnein:
## Summary: 

- Use Title Case for all headings in html report (capitalize the first letter of each major word)
- Adapt more to format of nextflow report (same table headers and value units)
- Ensure consistency between trace and html report 

#### Before: 
Screenshot 2025-08-06 at 11 46 40

#### After: 
Screenshot 2025-08-07 at 15 50 54

#### Before: 
Screenshot 2025-08-06 at 11 48 40

#### After: 
Screenshot 2025-08-07 at 15 52 45


nadnein:
I just fixed the `toReadableTimeUnits()` because it had some wrong behavior (there are now many tests to check if it's working correctly) and in the `CO2Record` I removed the `ciMarket` parameter as it should not be included in the tasks tables.

PR #286, opened 2025-08-11
Fix: Parsing a CSV file in `fromCsv`

nadnein:
## Fix CSV parsing in `fromCsv`

- Use `removeAt()` for `List` rows instead of `remove(value)` to avoid interpreting integers as indices (threw an error when reading in a custom CPU TDP file).
- Guard against missing `rowIndexColumn` (`findIndexOf` returning -1)  
- Use `remove(value)` for `LinkedHashSet` header instead of `removeAt`  
- Compare separator as `char` for reliability  
- Reset `escaped` flag per line to avoid quote-state leakage  
- Validate `rowIndexPos` before removing from row to prevent OOB errors

### DIFF
--- a/plugins/nf-co2footprint/src/main/nextflow/co2footprint/DataContainers/DataMatrix.groovy
+++ b/plugins/nf-co2footprint/src/main/nextflow/co2footprint/DataContainers/DataMatrix.groovy
@@ -125,6 +125,8 @@ class DataMatrix implements Matrix {
 
             // Add row to data
             data.add(row)
+            // Check if the line was properly closed
+            // If we are still escaped, it means the last quote was not closed
             assert escaped == false, "Unclosed quote in line: ${line}"
             start = 0
         }


plugins/nf-co2footprint/src/main/nextflow/co2footprint/DataContainers/DataMatrix.groovy
```diff
@@ -92,17 +95,18 @@ class DataMatrix implements Matrix {
...
         int start = 0
         int end = 0
-        char sepChar = (separator as String)[0] as char  // ensure char-vs-char comparison
+
+        boolean escaped = false // Track if we are inside a quoted field
 
         // Parse each line of the CSV, handling quoted fields and separators
         lines.each { line ->
-            boolean escaped = false       // ✅ reset per line
+            assert escaped == false, "Unclosed quote in line: ${line}"
```
JosuaCarl, line 103: This needs to come at the end of the loop to return the right line.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/DataContainers/DataMatrix.groovy
```diff
@@ -68,6 +68,9 @@ class DataMatrix implements Matrix {
     ) throws IOException {
         List<String> lines = Files.readAllLines(path)
 
+        // Ensure separator is a single character
+        char sepChar = separator?.charAt(0) ?: ','
```
JosuaCarl, line 72: Is it not, because `separator` is `char` which has to be one character?

nadnein: I commented on the last commit why I could not change the code as we agreed on: 
When I removed that line after defining the function parameter type as `Char` directly, I got an error because when you pass `','` to the function it is not automatically casted to `Char`. Therefore I now added this line at the beginning of the function to ensure everything works: 
`char sepChar = separator?.charAt(0) ?: ','`

plugins/nf-co2footprint/src/main/nextflow/co2footprint/DataContainers/DataMatrix.groovy
```diff
@@ -115,10 +125,11 @@ class DataMatrix implements Matrix {
 
             // Add row to data
             data.add(row)
+            assert escaped == false, "Unclosed quote in line: ${line}"
+            start = 0
```
JosuaCarl, line 129:
```suggestion

            // Check resetting of / reset text status trackers
            assert escaped == false, "Unclosed quote in line: ${line}"
            start = 0
```

JosuaCarl (approved):
:+1: 
Minor comment suggestion

PR #291, opened 2025-08-14
Fix/Refactor: Refactoring of converter, while addressing time conversion problems

### LINKED ISSUE #288: [Bug]: Memory uses the wrong metrics when requested `memory` is exceeded
### Describe the bug

The fallback of tasks exceeding their memory request is wrong.

### What happened?

The value that is used instead uses the maximum memory of the HEAD node.

### What should have happened?

Always using `memory` metric should suffice, because otherwise allocated memory can only be used, when no other task is currently using it (i.e. allocating free memory, but never taking memory from another task.

One could also make a case for using `peak_rss`, which tells us the maximum memory that was used during the task execution.

### Reproduction of the problem

Run a task that exceeds it's memory requirements and see the warning that is produced.

### In what computing environment did the problem occur?

- Plugin version: 1.0.0-rc.3

JosuaCarl:
## Related issues
- Closes #288 

## Related Pull-requests
- Inspired time-conversion adjustment from #284

## 🎯 Motivation
- Reduce the number of methods in Converter
- Fix time conversion inconsistencies
- Apply new structure to memory fallback error

## 📋 Summary of changes
- Added a `Quantity` class to track values along their unit and scale of that unit
- Combined scaling of Bytes and normal 1000-based unit system
- Simplified time conversion
- Fix of memory fallback with warnings on fallback

## ✅ Checklist


plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/Converter.groovy
```diff
@@ -32,52 +52,55 @@ class Converter {
...
-     * @param value Value that should be converted
-     * @param scope Symbol for scope of the unit (e.g. kilo = k)
-     * @param unit Name / symbol for the unit
-     * @return Converted String with appropriate scale
+     * @param value Value that should be converted (e.g. 10.1)
+     * @param scope Symbol for scope of the unit (e.g. kilo = k), default ''
+     * @param unit Name / symbol for the unit (e.g. B), default ''
+     * @param targetScale The scale that should be converted to (e.g. G)
+     * @param scalingFactor The factor with which to scale (e.g. 1024), default 1000
+     * @return Converted quantity with appropriate scale
```
nadnein, line 63:
```
/**
 * Converts a value to a human-readable string with appropriate unit scaling.
 *
 * @param value         Value to be converted (e.g. 10.1)
 * @param unit          Name or symbol for the unit (e.g. 'B'), default ''
 * @param scopes        List of scopes for the unit (e.g. ['', 'k', 'M', 'G', 'T'])
 * @param targetScope   The scope to convert to (e.g. 'G')
 * @return              Converted quantity as a string with the appropriate scale
 */
```
I think the function parameters could be simplified by removing the scalingFactor. And instead we could adapt the code and infer the scaling factor directly from the provided scopes (for example, SI units use 1000, binary units use 1024). This would make the function easier to use and understand and less confusing, since users would only need to specify the scopes and units, not the scaling logic. The current parameterization feels a bit too complex

nadnein (changes requested):

plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/Converter.groovy
```diff
@@ -1,29 +1,49 @@
...
- */ 
+ */
 @Slf4j
 class Converter {
 
+    /**
+     * Checks whether a element was found in a list and throws an error if not.
+     *
+     * @param unit Unit as a String
+     * @param units List of units that are available
```
nadnein, line 21:
```suggestion
     * @param element The element to search for
     * @param list The list in which to search
     * @return The index of the element in the list
```

plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/Converter.groovy
```diff
@@ -34,69 +54,72 @@ class Converter {
 
     /**
      * Converts a numeric value to a human-readable string with SI prefixes.
-     * For example, 1200 with unit 'Wh' becomes '1.2 KWh'.
-     * Scales the value up or down by factors of 1000 and adjusts the prefix accordingly.
+     * Scales the value up or down by a given factor and adjusts the scale prefix accordingly.
+     * The method is structured to take the quantity as in writing e.g. scaleUnits(1200, 'k', 'Wh', 'M'), with
+     * the last argument denoting the target to which the quantity is to be scaled.
+     * In this example, 1200 kWh is scaled to '1.2 MWh'.
```
nadnein, line 60: Nice 👍

plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/Converter.groovy
```diff
@@ -34,69 +54,72 @@ class Converter {
...
+     * Scales the value up or down by factors of 1000 and adjusts the prefix accordingly.
+     *
+     * @param value Value that should be converted
+     * @param scope Symbol for scope of the unit (e.g. kilo = k)
+     * @param unit Name / symbol for the unit
+     * @param targetScale Target scale to convert to
+     * @param precision Precision to round the value(s) to
+     * @return Converted String with appropriate scale and rounding
+     */
+    static String toReadableUnits(Double value, String scale='', String unit='', String targetScale=null, Integer precision=2) {
```
nadnein, line 101: Can't we just combine this with the `scaleUnits` to one method?

plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/Quantity.groovy
```diff
@@ -0,0 +1,85 @@
...
+
+import java.math.RoundingMode
+
+/**
+ * A number associated with a unit.
+ */
+class Quantity {
+    BigDecimal value
+    String unit
+    String scale
```
nadnein, line 11:
```suggestion
    String scale
    String unit
```

How could you!!! :astonished:

nadnein (changes requested):

plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/Converter.groovy
JosuaCarl: Yes, we could, but I wanted to separate the scaling and other stuff, like rounding. The `scaleUnits` method is further be used by the `CO2FootprintComputer` to scale values in a readable fashion. The key difference is, that this one returns a String, while the other still preserves the numeric representation. Therefore I would keep them separate.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/Quantity.groovy
JosuaCarl: The audacity. I'll immediately go flog myself and do the walk of shame 🔔

plugins/nf-co2footprint/src/main/nextflow/co2footprint/utils/Converter.groovy
```diff
@@ -34,69 +55,72 @@ class Converter {
...
+     * For example, 1200 with unit 'Wh' becomes '1.2 kWh'.
+     * Scales the value up or down by factors of 1000 and adjusts the prefix accordingly.
+     *
+     * @param value Value that should be converted
+     * @param scale Symbol for scale of the unit (e.g. kilo = k)
+     * @param unit Name / symbol for the unit
+     * @param targetScale Target scale to convert to
+     * @param precision Precision to round the value(s) to
+     * @return Converted String with appropriate scale and rounding
+     */
```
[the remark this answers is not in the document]
JosuaCarl: I only made small adjustment to this docstring, but modified the one at `scaleUnits` to make the difference clear.

PR #306, opened 2025-08-26
Add cpuPowerModel option, make config & computer more readable

nadnein:
### Add support for custom polynomial CPU power model (`cpuPowerModel`)  

This PR extends the CO₂ footprint plugin to support **user-defined CPU power models** as an alternative to static TDP values.  

#### Key changes:
- Added new config parameter `cpuPowerModel`:
  - Accepts a list of polynomial coefficients (`List<Double>` or `List<BigDecimal>`).
  - Interpreted as a polynomial in **core utilization (0–1)**, with highest degree first.  
  - Example:  
    ```groovy
    co2footprint {
      cpuPowerModel = [0.5, 10.0] // → P(coreUsage) = 0.5 * coreUsage + 10
    }
    ```
- If provided, this model **overrides TDP-based power draw lookups** for CPU cores.  
- New tests were added to cover the new functionality.
- Improved readability of `CO2FootprintComputer` and `CO2FootprintConfig`.
- Updated docs (+added figure) & CHANGELOG

#### Backward compatibility:
- If `cpuPowerModel` is not set, the plugin continues to use TDP-based power estimation as before.  
- All existing configs and workflows remain valid.


plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintComputer.groovy
```diff
@@ -209,4 +203,26 @@ class CO2FootprintComputer {
...
     }
+
+    /**
+    * Computes CPU power draw using the configured polynomial model.
+    *
+    * @param coeffs List of polynomial coefficients (highest degree first), as Double or BigDecimal.
+    * @param coreUsage CPU usage in percent (0–100).
+    * @return Estimated power draw [W/core], or null if no model configured.
+    */
+    BigDecimal getPowerDrawFromModel(def coeffs, BigDecimal coreUsage) {
```
JosuaCarl, line 214:
```suggestion
    BigDecimal getPowerDrawFromModel(List<BigDecimal> coeffs coeffs, BigDecimal coreUsage) {
```
I think List<Double> will then just be coerced to BigDecimal

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintComputer.groovy
```diff
@@ -209,4 +203,26 @@ class CO2FootprintComputer {
...
+        BigDecimal power = 0.0
+        int degree = coeffs.size() - 1
+
+        coeffs.eachWithIndex { c, i ->
+            power += (c as BigDecimal) * Math.pow(coreUsage.toDouble(), degree - i)
```
JosuaCarl, line 219:
```suggestion
        coeffs.eachWithIndex { BigDecimal c, int i ->
            power += c * coreUsage ** (degree - i)
```

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintConfig.groovy
```diff
@@ -92,6 +92,10 @@ class CO2FootprintConfig extends BaseConfig {
                 'machineType', 'Type of computer on which the workflow is run [\'local\', \'compute cluster\', \'\']',
                 null, String, Set.of(GString)
         )
+        defineParameter(
+                'cpuPowerModel', 'Polynomial coefficients for CPU power model (highest degree first)',
+                null, List, Set.of(Double, BigDecimal)
```
JosuaCarl, line 97:
```suggestion
                null, List, Set.of(List<Double>, List<BigDecimal>)
```

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintConfig.groovy
```diff
@@ -156,34 +163,48 @@ class CO2FootprintConfig extends BaseConfig {
...
-            }
-            else {
-                final String message = "machineType '${value('machineType')}' is not supported." +
-                        "Please chose one of ${supportedMachineTypes}."
+        /* ===== Determine CPU Power Model Parameters ===== */
+
+        // Use custom polynomial CPU power model if given
+        if (value('cpuPowerModel') != null) { 
+            if (value('cpuPowerModel').size() < 2) {
+                final String message = "cpuPowerModel must contain at least two coefficients (for a linear model)."
```
JosuaCarl, line 171: Why? We can also make one with one coefficent. This would then be equivalent to just giving the Powerdraw, but it works in theory.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintComputer.groovy
nadnein: I thought I have to leave it def when the values in the list can theoretically be both Double or BigDecimal, or not?

Issue #198, opened 2025-05-15
[Feature]: Export a machine readable standardized output format of the GHG records


JosuaCarl:
### Describe the feature

The exported file should follow some kind of international standard way of reporting the emissions to align the program with the FAIR principles of interoperability and reusability.

### Use case(s):

When other software wants to reliably access our output.

### Suggested implementation:

?

JosuaCarl:
A way to standardize the output would be to provide something similar to Nextflow's 
```config
trace {
    raw = true
}
```
for our trace file. Then the parameters would be machine-readable on the task level.

PR #313, opened 2025-10-02
Feature: Tree structure of results for machine-readable output

JosuaCarl:
## Related Issues
- Starts #198
  - The outpu behaviour does not change with this PR, but it lay the groundwork for doing so with ease
## 🎯 Motivation
- Provide a machine-readable file to enable easy parsing during analysis

## 📋 Summary of changes
- Passed on results in a hierarchical manner
- Removed accumulator in favor of tree structure
- Computed quantiles within plotly

## 📌 Important details
- Shortened Javascript though loop usage instead of multiple listings.
- This PR combines all trace values into `CO2Record`. The values can still be handled separately.
- Added Metric classes to fit purposes from `TraceRecord` and `CO2Record`
- Removed Converter class in favor of `Quantile`-class associated scaling

## ✅ Checklist

### DIFF
--- a/src/main/nextflow/co2footprint/CO2FootprintConfig.groovy
+++ b/src/main/nextflow/co2footprint/CO2FootprintConfig.groovy
@@ -31,7 +31,7 @@ import java.nio.file.Paths
  *     }
  *     report = {
  *       enabled: true,
- *       file: "co2footprint_report.txt
+ *       file: "co2footprint_report.html
  *     }
  *     ci = 300
  *     pue = 1.4
@@ -53,15 +53,15 @@ class CO2FootprintConfig extends BaseConfig {
         // Name, description, default value or function, return type, additional allowed types
         defineParameter(
                 'trace', 'Trace file config',
-                new FileSubConfig('trace', [:] as LinkedHashMap), FileSubConfig
+                new FileSubConfig([:] as LinkedHashMap, 'trace'), FileSubConfig
         )
         defineParameter(
                 'summary', 'Summary file config',
-                new FileSubConfig('summary', [:] as LinkedHashMap), FileSubConfig
+                new FileSubConfig([:] as LinkedHashMap, 'summary'), FileSubConfig
         )
         defineParameter(
                 'report', 'Report file config',
-                new FileSubConfig('report', [:] as LinkedHashMap), FileSubConfig
+                new FileSubConfig( [:] as LinkedHashMap, 'report', 'html'), FileSubConfig
         )
         defineParameter(
                 'location', 'Location GeoCode from Electricity maps',
@@ -132,11 +132,11 @@ class CO2FootprintConfig extends BaseConfig {
         configMap.each { name, value ->
             if (this.containsKey(name)) {
                 if (name == 'trace') {
-                    this.get('trace').set(new FileSubConfig('trace', value))
+                    this.get('trace').set(new FileSubConfig(value as Map, 'trace'))
                 } else if (name == 'summary') {
-                    this.get('summary').set(new FileSubConfig('summary', value))
+                    this.get('summary').set(new FileSubConfig(value as Map, 'summary'))
                 } else if (name == 'report') {
-                    this.get('report').set(new FileSubConfig('report', value))
+                    this.get('report').set(new FileSubConfig(value as Map, 'report', 'html'))
                 } else {
                     this.get(name).set(value)
                 }
--- a/src/main/nextflow/co2footprint/CO2FootprintObserver.groovy
+++ b/src/main/nextflow/co2footprint/CO2FootprintObserver.groovy
@@ -57,7 +57,6 @@ class CO2FootprintObserver implements TraceObserver {
 
     // Record for CI values during execution
     CiRecordCollector timeCiRecordCollector
-    CiRecordCollector getTimeCiRecordCollector() { timeCiRecordCollector }
 
     // Holds the the start time for tasks started/submitted but not yet completed
     @PackageScope
@@ -201,7 +200,7 @@ class CO2FootprintObserver implements TraceObserver {
         runningTasks.each { TaskId taskId, TraceRecord traceRecord -> aggregateRecords(traceRecord) }
 
         // Close all files (writes remaining tasks in the trace file)
-        traceFile.close(runningTasks)
+        traceFile?.close(runningTasks)
 
         // Finalize and aggregate all workflow statistics
         workflowStats.summarize()
@@ -221,8 +220,6 @@ class CO2FootprintObserver implements TraceObserver {
                 reportFile.write()
                 reportFile.close()
             }
-        }
-
         }
 
         // Close all files (writes remaining tasks in the trace file)
--- a/src/main/nextflow/co2footprint/Config/FileSubConfig.groovy
+++ b/src/main/nextflow/co2footprint/Config/FileSubConfig.groovy
@@ -17,6 +17,7 @@ import java.lang.Boolean
 class FileSubConfig extends BaseConfig {
     private String name
     private String timestamp = TraceHelper.launchTimestampFmt()
+    private String ending = 'txt'
 
     private void defineParameters() {
         // Name, description, default value or function, return type, additional allowed types
@@ -25,18 +26,20 @@ class FileSubConfig extends BaseConfig {
         )
         defineParameter(
                 'file', 'Path to the output file',
-                "co2footprint_${name}_${timestamp}.txt", String, Set.of(GString, Path)
+                "co2footprint_${name}_${timestamp}.${ending}", String, Set.of(GString, Path)
         )
     }
 
     /**
      * Load a nested map configuration and set up defaults and fallbacks.
      *
-     * @param subConfigName  Name for this sub-config - used to get a default file name for this output file
      * @param configMap      Map of configuration values (from Nextflow config)
+     * @param subConfigName  Name for this sub-config - used to get a default file name for this output file
+     * @param fileEnding     Ending type of the file
      */
-    FileSubConfig(String subConfigName, Map<String, Object> configMap) {
+    FileSubConfig(Map<String, Object> configMap, String subConfigName, String fileEnding='txt') {
         this.name = subConfigName
+        this.ending = fileEnding
 
         // Define the possible parameters of the configuration
         defineParameters()
--- a/src/test/nextflow/co2footprint/FileCreation/ReportFileCreatorTest.groovy
+++ b/src/test/nextflow/co2footprint/FileCreation/ReportFileCreatorTest.groovy
@@ -35,9 +35,9 @@ class ReportFileCreatorTest extends Specification{
     def setupSpec() {
         CO2FootprintConfig config = new CO2FootprintConfig(
                 [
-                        'traceFile': tempPath,
-                        'summaryFile': tempPath,
-                        'reportFile': reportPath,
+                        'trace': ['enabled': true, 'file': tempPath],
+                        'summary': ['enabled': true, 'file': tempPath],
+                        'report': ['enabled': true, 'file': reportPath],
                         'ci': 475.0
                 ],
                 Mock(TDPDataMatrix),
@@ -48,9 +48,9 @@ class ReportFileCreatorTest extends Specification{
         Session session = Mock(Session) {
             getConfig() >> [
                     co2footprint: [
-                            'traceFile': tempPath,
-                            'summaryFile': tempPath,
-                            'reportFile': reportPath,
+                            'trace': new FileSubConfig(['enabled': true, 'file': tempPath], 'trace'),
+                            'summary': new FileSubConfig(['enabled': true, 'file': tempPath], 'summary'),
+                            'report': new FileSubConfig(['enabled': true, 'file': reportPath], 'report', 'html'),
                             'ci': 475.0
                     ]
             ]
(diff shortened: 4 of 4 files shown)


plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintObserver.groovy
```diff
@@ -97,6 +90,9 @@ class CO2FootprintObserver implements TraceObserver {
         this.version = version
         this.config = config
 
+        // Collect metrics in hierarchical node structure
```
nadnein:
```suggestion
        // Create a RecordTree root node for the run, tagged as 'workflow',
        // to collect and organize execution metrics hierarchically.
```

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintObserver.groovy
```diff
@@ -68,12 +65,8 @@ class CO2FootprintObserver implements TraceObserver {
     @PackageScope
     Map<TaskId, TraceRecord> runningTasks = new ConcurrentHashMap<>()
 
-    // Stores CO₂ emission records by task ID
-    final private Map<TaskId, CO2Record> co2eRecords = new ConcurrentHashMap<>()
-    Map<TaskId,CO2Record> getCO2eRecords() { co2eRecords }
-
-    // Stores all trace records by task ID
-    final private Map<TaskId, TraceRecord> traceRecords = new ConcurrentHashMap<>()
+    // Stores all results and execution traces
```
nadnein:
```suggestion
    // Hierarchical tree that stores all results and execution traces
```

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintObserver.groovy
```diff
@@ -192,43 +181,34 @@ class CO2FootprintObserver implements TraceObserver {
...
 
-        // Close all files (writes remaining tasks in the trace file)
-        traceFile.close(runningTasks)
+        }
         summaryFile.close()
         reportFile.close()
+
+        log.info(
+            "🌱 The workflow run used ${workflowStats.value.getReadable('energy')} of electricity, " +
+            "causing ${workflowStats.value.getReadable('co2e')} of CO2 equivalents to be released into the athmosphere."
```
nadnein:
```suggestion
            "causing ${workflowStats.value.getReadable('co2e')} of CO2 equivalents to be released into the atmosphere."
```

plugins/nf-co2footprint/src/main/nextflow/co2footprint/Metrics/Calculator.groovy
```diff
@@ -0,0 +1,32 @@
+package nextflow.co2footprint.Metrics
+
+class Calculator {
+    static max(Object o1, Object o2) {
+        if (o1 > o2) {
+            return  o1
+        }
+        return o2
+    }
```
nadnein:
```suggestion
    /** Return the greater of two values. */
    static max(Object o1, Object o2) {
        return (o1 > o2) ? o1 : o2
    }
```

plugins/nf-co2footprint/src/main/nextflow/co2footprint/Metrics/Calculator.groovy
```diff
@@ -0,0 +1,32 @@
...
+
+    static Object add(Object o1, Object o2) {
+        if (o1 == null && o2 == null) { return null }
+        else if (o1 == null) { return o2 }
+        else if (o2 == null) { return o1 }
+        else { return o1 + o2 }
+    }
+
+    static Number weightedAverage(List<? extends Object> values, List<? extends Object> weights) {
```
nadnein:
```suggestion
     /**
     * Compute the weighted average of a list of values given a corresponding list of weights.
     * Null entries are removed before computation.
     *
     * @param values list of numeric values (nullable entries removed)
     * @param weights list of weights aligned with values (nullable entries removed)
     * @return weighted average as a Number, or null if inputs are empty
     */
    static Number weightedAverage(List<? extends Object> values, List<? extends Object> weights) {
```

plugins/nf-co2footprint/src/main/nextflow/co2footprint/Metrics/Metric.groovy
```diff
@@ -0,0 +1,26 @@
...
+     *
+     * @param value The value, saved in the metric
+     */
+    Metric(Object value, String type = value?.class?.getSimpleName(), String format = null) {
+        this.value = value
+        this.type = type
+        this.format = format
+    }
+
+    String getReadable() {
```
nadnein:
```suggestion
    /** Return the metric value as a human-readable string. */
    String getReadable() {
```

plugins/nf-co2footprint/src/main/nextflow/co2footprint/Metrics/Quantity.groovy
```diff
@@ -82,4 +83,8 @@ class Quantity {
 
         return readable
     }
+
+    Map<String, ? extends Object> toMap() {
```
nadnein:
```suggestion
    // Extend the parent map with scale and unit fields
    Map<String, ? extends Object> toMap() {
```

plugins/nf-co2footprint/src/main/nextflow/co2footprint/Records/CO2Record.groovy
```diff
@@ -138,18 +152,68 @@ class CO2Record extends TraceRecord {
...
+        }
+        return new CO2Record(store)
+    }
+
+    /**
+     * Create a map with all parameters in their "raw" and "readable" state. The "raw" value contains the unit and scope,
+     * as well as comments to contextualize the value.
+     *
+     * @param onlyCO2parameters
+     * @return
```
nadnein:
```suggestion
     * @param onlyCO2parameters if true, keep only CO₂-specific keys
     * @return map of field → { raw: {value, type}, readable: value }
```

plugins/nf-co2footprint/src/main/nextflow/co2footprint/ResultsTree/RecordTree.groovy
```diff
@@ -0,0 +1,204 @@
...
+
+    /**
+     * Summarizes the content of a Node by adding the sum of the children as a Map representation to its attributes.
+     * This methods works recursively by calling the summarize method with the assumption that the bottom layer is additive.
+     *
+     * Example:
+     * X( Y1(a = 1, b = 1), Y2(a = 1, b = 2) ).summarize()
+     * => X( a = 2, b = 3; Y1(...), Y2(...) )
+     *
+     * @return This RecordTree node with summary values
```
nadnein:
```suggestion
     * Example:
     * Parent(
     *   Child1(value = {a:1, b:1}),
     *   Child2(value = {a:1, b:2})
     * ).summarize()
     *
     * => Parent(value = {a:2, b:3}, children = [Child1(...), Child2(...)])
     *
     * @return this RecordTree node with aggregated summary values
```

plugins/nf-co2footprint/src/main/nextflow/co2footprint/ResultsTree/RecordTree.groovy
```diff
@@ -0,0 +1,204 @@
...
+import nextflow.co2footprint.Records.CO2Record
+
+/**
+ * A tree structure for CO2Records for collecting workflow results. The tree nodes are linked with a parent-children
+ * relationship.
+ */
+class RecordTree {
+    final name
+    final Map attributes
+    CO2Record value
```
nadnein: Can we rename those like this (Also all other occurences then of course)? I find these too generic and it also took me some time to understand the difference between value and attributes
```suggestion
    final Map summaryMetrics // aggregated attributes for any node
    CO2Record rawRecord         // only on leaf nodes (tasks)
```

nadnein (changes requested):
I requested a couple of changes. After those are resolved I will review the remaining part of the PR :)

plugins/nf-co2footprint/src/main/nextflow/co2footprint/CO2FootprintObserver.groovy
```diff
@@ -5,10 +5,10 @@ import groovy.util.logging.Slf4j
 
 import nextflow.Session
 import nextflow.co2footprint.Records.CO2Record
-import nextflow.co2footprint.Records.CO2RecordAggregator
 import nextflow.co2footprint.FileCreation.ReportFileCreator
 import nextflow.co2footprint.FileCreation.SummaryFileCreator
 import nextflow.co2footprint.FileCreation.TraceFileCreator
+import nextflow.co2footprint.ResultsTree.RecordTree
```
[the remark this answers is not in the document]
JosuaCarl: Yes, we can do both. I started with a Tree that had `TraceRecord` and `CO2Record` leaves and the naming stuck.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/Metrics/Quantity.groovy
JosuaCarl: That is not what the method does. It instead crates a Map from the object with all its attributes.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/Records/CO2Record.groovy
```diff
@@ -18,91 +20,92 @@ import nextflow.trace.TraceRecord
...
+             case 'memory' -> Converter.scaleUnits(value as Double, 'G', 'B', '').toMap()
+             default -> null
+         }
+         if (rawValue) { return rawValue }
+         return switch (FIELDS.get(key)) {
+            case 'date' -> new Quantity(value as Number, '1970-01-01T00:00:00Z -> x', 'ms').toMap()
+            case 'perc' -> new Quantity(value as Number, '%', '').toMap()
+            case 'mem' -> new Quantity(value as Number, '', 'B').toMap()
+            case 'num' -> new Quantity(value as Number, '', '').toMap()
+            default -> new Metric(value).toMap()
```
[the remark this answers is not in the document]
JosuaCarl: The `num` class/tag is present in the `TraceRecord`'s `FIELDS` map, which stores the type of the map entries. I think we don't catch all nums beforehand with the named variables, therefore it is needed.

The Metric class is uses for things, such as String/`str` variables. It is also needed to provide a gerneralized version for the `Quantity` class.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/Records/CO2Record.groovy
```diff
@@ -138,18 +152,68 @@ class CO2Record extends TraceRecord {
...
-    List<String> getReadableEntries(List<String> order=store.keySet() as List) {
+    List<String> getReadableEntries(List<String> order=co2Keys as List) {
         return order.collect { String key -> getReadable(key) }
     }
 
     /**
-     * Renders the JSON output of a CO2Record.
+     * Add an element from one CO2Record to the same in another.
+     * This can correspond to a simple addition, but also cover non-addable value such as Strings, which are collectes,
+     * or accumulation with a weighted average.
```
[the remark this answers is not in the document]
JosuaCarl: The description does not fit what the method does. It does only combine the specified metric from the two `CO2Record` instances, not whole records.

plugins/nf-co2footprint/src/main/nextflow/co2footprint/ResultsTree/RecordTree.groovy
JosuaCarl: I would rename the `value` field, but leave the attributes field, as this corresponds to the XML naming scheme. Also when `.summarize()` was called, all `value` entries are filled. Only at the start only the leaves hold the values.

JosuaCarl: Renamed values: `attributes` -> `metaData`, `value` -> `co2Record`

Does this capture the spirit of your intention?

plugins/nf-co2footprint/src/main/nextflow/co2footprint/Metrics/Metric.groovy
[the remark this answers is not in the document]
JosuaCarl: I would argue that the naming would provide a distinction between metrics, which include all variables, like names and models,... and `Quantities` which only include numeric metrics.

From a technical standpoint we could integrate both classes, but I don't find it very intuitive. We would need additional checks in the `Coverter` class.

src/main/nextflow/co2footprint/FileCreation/TraceFileCreator.groovy
```diff
@@ -91,10 +74,10 @@ class TraceFileCreator extends BaseFileCreator {
      * @param traceRecord   TraceRecord for the task
      * @param co2Record     CO2Record for the task
      */
-    void write(TraceRecord traceRecord, CO2Record co2Record){
+    void write(CO2Record co2Record){
         if (!created) { return }
 
-        List<String> recordedEntries = entryKeys.collect { String key -> getReadableEntry(key, traceRecord, co2Record) }
+        List<String> recordedEntries = co2Record.getReadableEntries(entryKeys) //  entryKeys.collect { String key -> getReadableEntry(key, traceRecord, co2Record) } ?
```
nadnein: Should the comment here stay there?

src/main/nextflow/co2footprint/Logging/CustomCaptureAppender.groovy
```diff
@@ -52,9 +51,9 @@ class CustomCaptureAppender extends AppenderBase<ILoggingEvent> {
             }
             else {
                 // Choose appropriate level
-                switch (event.level) {
-                    case Level.ERROR -> renderer.appendError(message)
-                    case Level.WARN -> renderer.appendWarning(message)
+                switch (event.level.levelInt) {
```
nadnein: What is this `levelInt` good for?

src/main/nextflow/co2footprint/Metrics/Quantity.groovy
```diff
@@ -47,7 +56,7 @@ class Quantity {
      */
     Quantity round(Integer precision=2, RoundingMode roundingMode=RoundingMode.HALF_UP) {
```
nadnein: Maybe just use RoundingMode.HALF_UP in the function body as in `floor`? WHy is it a function parameter?

nadnein (approved):

PR #315, opened 2025-10-13
Refactor: switch from the full plotly.min.js to plotljs-cartesian

mashehu:
## 🎯 Motivation
- make the footprint of the HTML report smaller 😉 

## 📋 Summary of changes
- switched plotlyjs to the custom bundle of plotly only containing scatter and box plot functions. 
	- The bundle file was generated following instructions in [plotly/plotly.js@master/CUSTOM_BUNDLE.md](https://github.com/plotly/plotly.js/blob/master/CUSTOM_BUNDLE.md) and the command `npm run custom-bundle -- --traces box,scatter`
  - This reduces the file size of plotly from 3.4MB to 1.1MB (which is 0.3MB smaller than the pre-bundled [plotlyjs-cartesion](https://github.com/plotly/plotly.js/blob/master/dist/README.md)

## ✅ Checklist


JosuaCarl (approved):
Looks good to me :+1:

I think you managed to push down the size of the Reports from ~3.6 to ~1.6MB, thanks a lot.

PR #316, opened 2025-10-14
fix: hard code locale for decimals

mashehu:
## 🎯 Motivation
If the java locale is set to a country that uses `,` as a decimal separator (e.g. Germany or Sweden), the report would be wrong in other countries.

## 📋 Summary of changes
hard-coded the locale to US to enforce `.` as a decimal separator.

## ✅ Checklist


JosuaCarl (approved):
Good catch :+1:

PR #318, opened 2025-10-16
Update CPU TDP dataset and stabilize tests with fixed test CSV

nadnein:
### Description
- Updated `CPU_TDP.csv` with the latest data.  
- Added a fixed test dataset `CPU_TDP_test.csv` under `src/testResources` 
- Adapted tests to use the fixed dataset to avoid updating too many tests when cpu tdp data table is updated


JosuaCarl (changes requested):
Minor comment

PR #324, opened 2025-11-06
Feature: Separate reporting energy consumption

JosuaCarl:
## 🎯 Motivation
- Report intermediary energy steps in the trace file

## 📋 Summary of changes
- Added `rawEnergyProcessor`, `rawEnergyMemory` to `CPURecord` and trace file

## ✅ Checklist


JosuaCarl:
I think that you should be able to infer the way all values are derived in the trace file. The CPU should also be mostly the same for every task, but we print it anyway so our calculations are retraceable. And the PUE is so far nowhere reported, but changes the final values by quite a lot.

skrakau:
It should be already reported in the HTML reported in the section of used parameters and in the summary file, if not it's a bug.

The same would apply for other such values, and in any case I think this should be done consistently. But we can discuss the general strategy.

JosuaCarl:
You are right, it is in the HTML file. I would still like to include the PUE because it's a very high-level component that changes the energy-consumption quite a lot, depending on the efficiency.

PR #363, opened 2026-03-25
Fix: Resolve test resource paths

nadnein:
## Problem

`FileChecker` resolved its working directory via `this.class.getResource()`, which points to `build/resources/test/` (the compiled classpath). This caused:
- Changes to `file_checks.json` requiring a rebuild to take effect.
- Failed snapshot files being written into `src/testResources/` instead of `build/`.

## Changes

- `checksDirectory` now resolves from `src/testResources/` so tests always read directly from the source tree.
- Write locations (`recordPath`, `failPath`) remain under `build/resources/test/`.

### Refactoring (no behaviour change)
- Introduce `buildChecksDirectory` field to avoid repeating the `user.dir` lookup.
- Rename `newPath` → `failedSnapshotPath` for clarity.
- Reorder class fields into logical groups.

## Testing

`./gradlew check` ✅

### DIFF
--- a/src/test/nextflow/co2footprint/TestHelpers/FileChecker.groovy
+++ b/src/test/nextflow/co2footprint/TestHelpers/FileChecker.groovy
@@ -31,8 +31,16 @@ class FileChecker {
     // Directory with files to check
     private Path checksDirectory
 
+    // Directory with recorded files in build test resources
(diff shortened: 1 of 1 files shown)


JosuaCarl (approved):
:+1:

PR #370, opened 2026-04-10
Align plot height behavior in the HTML report

nadnein:
## Description

This PR makes the per-process emissions plot and the swimlane plot use the same height logic in the HTML report.

Both plots now scale with the number of visible process rows and share the same cap:

`Math.max(200, Math.min(900, 80 + rowCount * 42))`

This keeps the charts more consistent and avoids excessively tall plots when many processes are shown.


JosuaCarl (approved):
Looks good to me.

PR #375, opened 2026-04-28
Feature: Use provenance file extension run

### LINKED ISSUE #198: [Feature]: Export a machine readable standardized output format of the GHG records

JosuaCarl:
## 🎯 Motivation
- Enable post-run with provenance file in order to cover all results
- Add all necessary metrics to reproduce the calculation to trace and provenance file
- Mitigate floating point errors (when reproducing calculations)
  - Was result of incorrect `realtime` saving due to conversion h <-> ms

## 📋 Summary of changes
- Added ability to parse provenance file into a `CO2RecordTree`

## 📌 Important details
- Replaced `Double` numbers with `BigDecimal` in most cases
- Optimized imports for clearer structure and removal of redundant classes
- Consolidated post-run features into `CO2FootprintExtension` to avoid duplicate functionality
- Added post-run detection to `CO2FootprintCalculator` to enable access to previously assigned values, if they are not explicitly assigned again
- Moved test files for post-run into cli subfolder
- Saved `realtime` as its real ms value to avoid rounding errors

## ✅ Checklist


docs/usage/extension.md
```diff
@@ -69,8 +69,23 @@ workflow {
 ###### `parseTraceFile`
 Parse the trace file into a list of TraceRecord instances. Accepts the [shared arguments](#shared-arguments).
 
+- **tracePath**:
```
nadnein, line 71: When there is now the filePath argument, which can read in both, the trace and provenance file, this is redundant(?)

docs/usage/output.md
```diff
@@ -16,10 +16,22 @@ The nf-co2footprint plugin creates three output files:
...
+    - The value accumulation from `tasks` over `process` to `workflow` level happens differently for different values:
+        - Weighted average
+          - In relation to energy: `carbon_intensity`, `powerdraw_cpu`, `carbon_intensity_market`, `powerdraw_memory`
+          - In relation to runtime: `%cpu`, `%mem`, `vmem`, `rss`, `cpus`, `pue`
+        - Maximum: `memory`, `cpus`, `complete`, `attempt`, `peak_vmem`, `peak_rss`
+        - Minimum: `submit`, `start`
+        - Accumulation in set: `cpu_model`, `status`, `name`, `cpu_power_model`
+        - All other values are added up
+  
+    We hope that the accumulation style provides sensible values at every level and valuable insights for you. You can always suggest changes at our [Github issue page](https://github.com/nextflow-io/nf-co2footprint/issues).
```
nadnein, line 34:
```suggestion

```
I think this is unnecessary, as this is clear anyways

docs/usage/extension.md
nadnein, line 45: change to filePath

src/main/nextflow/co2footprint/FileCreation/ProvenanceFileCreator.groovy
```diff
@@ -153,4 +177,88 @@ class ProvenanceFileCreator extends BaseFileCreator {
...
+        Map<String, Object> jsonMap = new JsonSlurper().parseText(jsonString) as Map<String, Object>
+        return readJsonLd(jsonMap)
+    }
+
+    /**
+     * Recursively read a JSON-LD map and transform it back to the original CO2RecordTree map structure.
+     * This is the inverse of transformToJsonLd, and assumes the same JSON-LD structure as produced by that function.
+     *
+     * @param treeMap The JSON-LD map to transform back to the original structure
+     * @param root Whether this is the root level of the tree, used to determine the level of the root node
```
nadnein: I think `isRoot` would be a better variable name? (But not so important)

src/main/nextflow/co2footprint/CO2FootprintCalculator.groovy
```diff
@@ -54,16 +55,18 @@ class CO2FootprintCalculator {
     *
     * @param trace   The TraceRecord containing task resource usage.
     * @param timeCiRecords Collector for carbon intensity records.
+    * @param postRun Whether the computation is performed after the run (true) or during the run (false). Used to determine whether previous values are used if not set explicitly otherwise.
```
nadnein: Also here maybe `isPostRun`for a boolean variable as name?

src/main/nextflow/co2footprint/CO2FootprintCalculator.groovy
```diff
@@ -207,4 +238,26 @@ class CO2FootprintCalculator {
...
+
+    /**
+     * Utility method to use the from the trace value from a previous run if the entry was not explicitly set in the config.
+     *
+     * @param config The CO2FootprintConfig object containing the configuration for the current run
+     * @param configKeys The entry keys in the config that trigger the value to be set explicitly
+     * @param configValue The current config value
+     * @param trace TraceRecord or CO2Record, potentially with values from previous runs
+     * @param traceKey The entry key in the trace
+     * @return The value from the trace if the config entry was not explicitly set and the trace contains the value, otherwise the current config value
```
nadnein: I don't understand this function completely, could you maybe explain it in more detail like in post run setting this and this is used and in live run this is used or so? And why do we need the first and second param? Isn't the third enough? And the postRun param is missing in the description

nadnein (changes requested):
Crazy work! I have some notes

docs/usage/output.md
JosuaCarl: I just wanted to restate this here, because the accumulation of values is very choice dependent and often has no right answer.

JosuaCarl: Take for instance the accumulation of `memory` it could be summed to know how much the individual tasks requested in total (but we don't know whether they ran in parallel), we could use the time average, to know what was requested on average by the tasks dependent on how long they ran (but that ignores parallelism), and we could take the maximum (but that could still be too little because of parallelism).

The point is that there is no right answer here and I just wanted others to see that suggestions are welcome. Like with the missing CPU warning that tells you to report it so the list may grow.

docs/usage/extension.md
JosuaCarl: No, that is correct in this instance because we are using the `trace` mode implicitly.

src/main/nextflow/co2footprint/CO2FootprintCalculator.groovy
JosuaCarl: Some values are passed via the config (like lets say `pue`). When we want to replicate a run, it should calculate the exact same results if no parameters are changed, right? But if we just use the value given by the `config` in the `Calculator`, we could end up using the default/fallback instead of a previously set parameter.

So we use previous values if all of the following are true:
1. We are in a post run setting
2. The value was not explicitly specified in the config
3. The value is in `TraceRecord`

JosuaCarl: I have added this to the docstring

src/main/nextflow/co2footprint/CO2FootprintCalculator.groovy
```diff
@@ -207,4 +238,29 @@ class CO2FootprintCalculator {
...
+     * 2. The entry was not explicitly set in the config
+     * 3. The entry is in the `TraceRecord`
+     *
+     * @param config The CO2FootprintConfig object containing the configuration for the current run
+     * @param configKeys The entry keys in the config that trigger the value to be set explicitly
+     * @param configValue The current config value
+     * @param trace TraceRecord or CO2Record, potentially with values from previous runs
+     * @param traceKey The entry key in the trace
+     * @return The value from the trace if the config entry was not explicitly set and the trace contains the value, otherwise the current config value
+     */
```
nadnein, line 264:
```suggestion
    /**
     * Resolves a calculation parameter by choosing between the current config and a value
     * stored in a previous run's trace output.
     *
     * In post-run mode, the input trace may be a CO2Record from a prior plugin run that already
     * contains the parameter values used in the original calculation (e.g. PUE, carbon intensity).
     * This method preserves those original values — unless the user explicitly overrides them
     * in the current config.
     *
     * In normal (integrated) mode, always returns the current config value.
     *
     * Example: a prior run used PUE=1.4, stored in the trace. If the user re-runs post-run
     * estimation without setting 'pue' in config, this returns 1.4 from the trace. If the user
     * explicitly sets pue=1.2 in config, this returns 1.2.
     *
     * @param config      Plugin configuration for the current run
     * @param configKeys  Config keys that, if explicitly set by the user, force using the config value
     * @param configValue The value from the current config (used when not in post-run, or when overridden)
     * @param trace       The input trace record, potentially containing values from a previous run
     * @param traceKey    The field name in the trace to look up the previous value
     * @param isPostRun   Whether we are in post-run mode
     * @return The previous run's value if in post-run mode and not overridden, otherwise the config value
     */
```

nadnein (approved):

PR #376, opened 2026-04-28
Fix post-run metadata, zero-value fallback, CI weighting, and provenance/AWS edge cases

nadnein:
**Summary**

This PR fixes a handful of correctness bugs in the CO2 calculation and post-run reporting paths.

**What changed**

1. Fixed `getTraceOrDefault()` so real `0` / `0.0` trace values are preserved instead of being treated as missing and replaced by defaults.

2. Fixed workflow metadata generation in both the CLI and extension post-run paths to use the latest task `complete` timestamp, so workflow completion time and duration are calculated correctly.

3. Fixed time-weighted carbon intensity calculation to weight CI values by the duration they cover, which makes results correct for uneven CI sampling intervals.

4. Fixed AWS region normalization so values returned from `aws configure get region` are trimmed before lookup.

5. Fixed provenance JSON-LD output to emit `schema:DateTime` instead of the invalid `schema:DataTime`.

6. Updated and added focused regression tests for:
   - zero-value fallback behavior
   - uneven CI intervals
   - corrected CLI post-run workflow timing
   - provenance `DateTime` serialization

**Why**

These issues could lead to incorrect defaults being applied, wrong workflow duration in post-run reports, biased CI calculations when timestamps are unevenly spaced, failed AWS region matching due to trailing newlines, and invalid JSON-LD typing in provenance output.

**Validation**

Validated with focused Gradle test runs covering:
- `CO2FootprintCalculatorTest`
- `CO2FootprintCLITest`
- `CiRecordCollectorTest`
- `AWSRegionsDataMatrixTests`
- `ProvenanceFileCreatorTest`

### DIFF
--- a/src/main/nextflow/co2footprint/Records/CiRecordCollector.groovy
+++ b/src/main/nextflow/co2footprint/Records/CiRecordCollector.groovy
@@ -132,8 +132,7 @@ class CiRecordCollector {
         }
 
         if (!timestamps) {
-            String message = "Found no timestamps in timeCIs '${timeCIs}' that are " +
-                    "before, during or after the frame of the traceRecord: '${trace}' (${start}-${end})."
+            String message = "No carbon intensity timestamps are available in timeCIs for traceRecord '${trace}' (${start}-${end})."
             log.error(message)
             throw new MissingValueException(message)
         }
--- a/src/main/nextflow/co2footprint/Records/CiRecordCollector.groovy
+++ b/src/main/nextflow/co2footprint/Records/CiRecordCollector.groovy
@@ -122,15 +122,7 @@ class CiRecordCollector {
                 ZoneId.systemDefault()
         )
         Long duration = start.until(end, ChronoUnit.MILLIS)
-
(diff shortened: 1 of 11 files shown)


src/main/nextflow/co2footprint/Records/CiRecordCollector.groovy
```diff
@@ -139,40 +131,32 @@ class CiRecordCollector {
...
-            averageCi += timeCIs.get(last) * getWeight(last, end)
-        }
-        else if (before) {
-            averageCi += timeCIs.get(before.max()) * getWeight(start, end)
-        }
-        else if (after) {
-            averageCi += timeCIs.get(after.min()) * getWeight(start, end)
-        }
-        else {
+        if (!timestamps) {
```
JosuaCarl, line 134: Can you change the corresponding error message to "No timestamps found, or somthing like that"?

JosuaCarl (changes requested):

JosuaCarl (approved):
:+1:

PR #383, opened 2026-05-04
Fix floating point error and refactor Bytes for value types

JosuaCarl:
## Related Pull-requests
- #382 fixes plugin test error that might come up during testing

## 🎯 Motivation
- Use decimal based division to convert between Bytes, as this is in accordance with SI standards
- Fix a floating point division error for big numbers

## 📋 Summary of changes
- Renaming to improve verbosity of code

## ✅ Checklist


nadnein (changes requested):
Nice thanks!
One small comment, I think we should mention in the docs somewhere that we use the (correct) 10^9 scaling factor for Bytes --> GB conversion and Nextflow does not and therefore it could happen that e.g. 12 GB set as memory by the user for a process is reported as 12.88 GB (SI) in the plugin output files.

nadnein (approved):

PR #385, opened 2026-05-04
Fix: zero division crashes when task realtime is 0

nadnein:
When a Nextflow task reports `realtime = 0`, two separate code paths crash with `ArithmeticException`:

- `Calculator.weightedAverage()` divides by the sum of weights, which is 0 when all realtime values are 0 — added a norm == 0 guard returning null early.
- `CiRecordCollector.getWeightedCI() `divides by duration (= complete - start), which is 0 when both timestamps are identical — added an early return of the CI at the start timestamp.


JosuaCarl (approved):
:+1:

PR #397, opened 2026-06-18
Fix/session trace accumulation

### LINKED ISSUE #396: [Bug]: Session memory does not encompass task memory, but suggests so
The `memory` value on the session level is the heap size of the JVM. But some processes that are children of the JVM can allocate memory outside of it by accessing previously running engines (like Slurm, AWS Batch, Apptainer,...).

Therefore, the `%cpu` and other values are only accurate, if the processes run on the same machine an `memory`, `peak_rss` and `vmem` are inaccurate in almost any case.

Solution:
- Session should be an extra branch in the provenance file that is on the same level as "workflow" and adds up to a "Run" value with all the info from the tasks.
- The session value must be cut off at the point where Nextflow session tracking comes in to avoid duplicates

JosuaCarl:
## Related Issues
- Closes #396

## 🎯 Motivation
- Limit the scope of the session trace to the head job and construct session value from addition

## 📋 Summary of changes
- Renamed Session values to HeadJob
- Excluded tasks from memory and CPU tracking on Session level

## ✅ Checklist


docs/usage/output.md
```diff
@@ -17,12 +17,12 @@ The nf-co2footprint plugin creates three output files:
...
 
 - **`provenance`**
-  The provenance file contains all trace information contributing to the emission calculation in a tree structure with the levels in descending order `session -> workflow -> process -> task`. Example: A workflow consists of multiple processes / has multiple `process` level children.
+  The provenance file contains all trace information contributing to the emission calculation in a tree structure with the levels in descending order `session -> head job & workflow -> process -> task`. Example: A workflow consists of multiple processes / has multiple `process` level children.
   The file design adheres to the javascript object notation linked-data (JSON-LD) format with type context definitions from schema.org and bioschemas.org.
-  The total session emission estimation includes everything from the point of plugin start to stop in a similar manner to how Nextflow defines a `TraceRecord`, but through the Java-native OSHI library.
+  The head job emission estimation includes all processes except tasks from the point of plugin start to stop in a similar manner to how Nextflow defines a `TraceRecord`, but through the Java-native OSHI library.
```
nadnein, line 22: I think we should be more concrete and also mention that not allocated memory is used but peak_rss and that its just added to the sum of the task specific resource usages recorded by nextflow

nadnein (changes requested):

nadnein (approved):

PR #398, opened 2026-06-18
Fix realtime display bugs in reports and trace output

nadnein:
## Summary
- **Fix realtime column in HTML report** showing wrong values: the DataTable column was reading the `time` field (requested time limit) instead of `realtime` (actual wall-clock runtime), causing "always 1h" in integrated mode and "-" in post-run mode.
- **Fix floating point rounding artifacts** in duration formatting (e.g., "2s 1000ms" instead of "3s", "2s 0ms" instead of "2s") by eliminating the lossy `ms → hours → ms` round-trip conversion. The realtime value is now passed directly in milliseconds from the trace to the CO2Record, while `runtime_h` is computed separately only for the energy calculation where it's needed.
- **Consolidate extension point docs** to reduce repetition between CLI and function argument descriptions by extracting shared arguments (`tracePath`, `delimiter`) into a common section.

## ✅ Checklist


JosuaCarl:
Soo, the elimination of the round trip conversion and the consolidation of common function in `CO2FootprintCLI` and `CO2FootprintExtension` are already part of #375.

Regarding the first: I also added a lot more to eliminate floating point errors in said PR.

Regarding the second: I don't see any changed docs nor changes in the said classes. Is this already pushed?

I would prefer that we'd merge the older PRs first and then address this one. This again reflects the problem of leaving PRs hanging around.

PR #405, opened 2026-07-09
Fix realtime column in html report

nadnein:
- Fixed `realtime` DataTable column reading `time` (requested limit) instead of `realtime` (actual runtime) — caused "always 1h" / "-" display bugs
- Updated `report_test.html` fixtures + checksums to match


JosuaCarl (approved):
:+1:

PR #408, opened 2026-07-10
Deduplicate log warnings across all CO2 estimation entry points

nadnein:
## Summary
- `CO2FootprintCLI.postRun()` and the `calculateCO2`/`parseTraceFile` function extension never registered the log deduplication filter, so warnings like "unknown CPU model" were logged in full for every task instead of collapsing to one WARN + DEBUG `[DUPLICATE]`s, as happens in the live observer path.
- Both entry points now call `adaptLogging()`, which is now idempotent so it's safe to call from multiple places.

Before it looked like this: 
Screenshot 2026-07-09 at 13 25 58


JosuaCarl (approved):
Looks good :+1:

PR #416, opened 2026-07-28
Feature: Added CI records to provenance file

### LINKED ISSUE #414: [Feature]: Add CI timeline to provenance file
### Describe the feature

`{"2026-07-15T13:00":192,"2026-07-15T16:00":215,"2026-07-15T09:00":301,"2026-07-15T11:00":258,"2026-07-15T10:00":270,"2026-07-15T17:00":272,"2026-07-15T12:00":228,"2026-07-15T15:00":171,"2026-07-15T14:00":182};` is in the report, but should also be preserved in the provenance file.

### Use case(s):

For re-plotting.

### Suggested implementation:

_No response_

JosuaCarl:
## 🎯 Motivation
- Add information on used CI Records to provenance file

## 📋 Summary of changes
- Changed provenance file creation to incorporate CI Records

## 📌 Important details
- Test: Adjusted integration test config

## ✅ Checklist


nadnein:
It stores only the time-weighted average CI value for a task if the API was used right? Shouldn't the provenance file store the raw data, like the series of CI's (and also corresponding durations somewhere) if more than 1 CI influenced a task's emissions?

JosuaCarl:
> It stores only the time-weighted average CI value for a task if the API was used right? Shouldn't the provenance file store the raw data, like the series of CI's (and also corresponding durations somewhere) if more than 1 CI influenced a task's emissions?

That was also done previously. Now the CI raw data is also stored, but only at the topmost level in `carbon_intensity_records`, so at `session` or `workflow`. I found it redundant to report them multiple times. But we could also change that.

src/main/nextflow/co2footprint/FileCreation/ProvenanceFileCreator.groovy
```diff
@@ -70,6 +81,25 @@ class ProvenanceFileCreator extends BaseFileCreator {
...
+    private static Map<String, Object> ciRecordsToJsonLd(Map<Instant, Number> ciRecords ) {
+        Map<String, Object> ciRecordsMap = ['@type': 'schema:ItemList']
+        ciRecords.eachWithIndex { Instant instant, Number ci, Integer i ->
+            ciRecordsMap.put( 'itemListElement', 
+                [ '@type': 'schema:ListItem', position: i, item: [
+                    '@type': 'schema:Observation', observationDate: instant.toString(), value: ci, unitText: 'g/kWh'
+                    ]
+                ]
+            )
+        }
```
nadnein, line 117:
```suggestion
        List<Map<String, Object>> itemListElement = []
        ciRecords.eachWithIndex { Instant instant, Number ci, Integer i ->
            ciRecordsMap.add( 'itemListElement', 
                [ '@type': 'schema:ListItem', position: i, item: [
                    '@type': 'schema:Observation', observationDate: instant.toString(), value: ci, unitText: 'g/kWh'
                    ]
                ]
            )
        }
        ciRecordsMap.put('itemListElement', itemListElement)
```
Claude told me this is a bug as put on a Map key overwrites the previous value and it doesn't append. Meaning that in every iteration the previous value is replaced by a new one

nadnein (changes requested):

src/main/nextflow/co2footprint/FileCreation/ProvenanceFileCreator.groovy
JosuaCarl: The suggestion did not solve the problem, but the analysis was correct. Did not register with me because I only tested small workflows.

nadnein (approved):
looks good now 👍

PR #417, opened 2026-08-07
Fix report plot/table resizing, SVG-friendly titles, and plot export quality

nadnein:
### Summary
- "Per process emissions" plot and the options table now resize with the browser window like the rest of the report 
- Plot titles (CI/energy, swimlane, memory optimization) are now HTML headings above each plot instead of baked into the Plotly canvas, so downloaded images/SVGs only contain the chart.
- Plot downloads now export as SVG at the plot's current on-screen size instead of a fixed-size rasterized PNG.
- Adds a `provenance.includeNulls` config option (default false) to optionally keep null-valued fields (e.g. cpu_model when detection fails) in the provenance JSON. This also fixes a latent NPE in ProvenanceFileCreator that only nulls could trigger, and stabilizes the integration test's provenance line-count check by enabling the option for that test.
- Updated docs


src/test/CO2PluginFullTest.groovy
```diff
@@ -96,8 +96,14 @@ class CO2PluginFullTest extends Specification {
...
         fileChecker.compareNumLines(tracePath, 8)
         fileChecker.compareNumLines(summaryPath, 30)
-        fileChecker.compareNumLines(reportPath, 1842)
-        fileChecker.compareNumLines(dataPath, 4599)
+        fileChecker.compareNumLines(reportPath, 1851)
+
+        // The provenance JSON's line count depends on details of the live
+        // pipeline run (e.g. observed 4592 on macOS/Docker Desktop locally
+        // vs. 4599 on GitHub Actions' Linux runners for the same code), so
+        // an exact match isn't reliable across environments. Allow a margin.
```
JosuaCarl: Why? Do you have an idea? Is there information missing for one of the runs? Did you compare both files? This has never been an issue before.

JosuaCarl (commented):
Can you compare the two generated data files and explain why there is a difference in the number of lines?

src/test/CO2PluginFullTest.groovy
```diff
@@ -96,8 +96,12 @@ class CO2PluginFullTest extends Specification {
         // Check line count
         fileChecker.compareNumLines(tracePath, 8)
         fileChecker.compareNumLines(summaryPath, 30)
-        fileChecker.compareNumLines(reportPath, 1842)
-        fileChecker.compareNumLines(dataPath, 4599)
+        fileChecker.compareNumLines(reportPath, 1851)
+
+        // includeNulls=true (see integration/nextflow.config) keeps entries
```
JosuaCarl: I would remove this comment, because it is mostly relevant to this PR.

src/testResources/integration/nextflow.config
```diff
@@ -31,6 +31,10 @@ co2footprint {
         file = "${params.outdir}/pipeline_info/data_test.json"
         enabled = true
         emissionMetricsOnly = false
+        // Keep null-valued entries (e.g. cpu_model, when detection flakes on a
```
JosuaCarl: I would shorten it to "`includeNulls` ensures stable line count on different machines".

JosuaCarl (approved):
Only a few remarks to two comments which are a bit verbose for a small test fix.

Looks good otherwise :+1:

